PR #4, opened 2024-01-24
Fix/org avatar rendering

### LINKED ISSUE #1: Unreliable Organization Avatar Rendering
There's a bug where the organization avatar images do not reliably display on each tile. The code is a bit messy around the state management of this so I expect this to be a pretty tricky problem to solve.

For me, when I first open up the UI, when the tasks are loaded from the cache they never render the images. However when I login with GitHub, they do render:

work ubq fi_

I remember the original obstacle was related to rate limits (when not logged in, can only make something like 60 requests in a minute; this is problematic when there are 30+ issues (issue contents, and organization avatar is already two requests.) 

I tried implementing a cache system for issues (done, and well implemented) as well as the avatars (not well implemented.)

0x4007:
Resolves #1

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,19 @@
+name: Build
+
+on:
+  push:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js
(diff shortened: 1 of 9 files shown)


0x4007:
Pretty epic ChatGPT review I'm implementing right now. https://chat.openai.com/c/cd5866dd-6d3a-42c0-97ee-37b1e414e958

PR #7, opened 2024-01-29
Fix/caching

### LINKED ISSUE #6: New Cache Schema
###### Draft

There's several properties we must associate with issues in the context of displaying them correctly. 
1. `origin` - where the issue was loaded from `localStorage | octokit`
   - Any `octokit` origin issue (freshly loaded from the API) should have a visual indicator as being "new"
2. `fullIssueId` - we only load issue previews from `@ubiquity/devpool-directory/issues` and we need to associate the preview issue ID to the "full" issue ID that it points to. 

- I also think that we should immediately display from cache first, and then fetch updates in the background. 
- If an issue is no longer present according to octokit, we need to sync our localStorage (remove the issue) so that it does not display anymore.

0x4007:
Cache system refined for GitHub issues.


0x4007:
Just realized that this broke the avatar rendering on the GitHub issues.

PR #17, opened 2024-02-26
feat: cypress testing

### LINKED ISSUE #14: Stale data is displayed without a full cache clean
It seems that https://work.ubq.fi/ displays tickets that are closed, updated, or moved, regardless of their real state in the GitHub repo. This can be fixed by emptying the browser cache and reloading the page, which is an abnormal behavior.

Expected behavior:
The list of available bounties should reflect the available issues in the corresponding GitHub repositories.

gentlementlegen:
Resolves #14


cypress/fixtures/issue-1.json
0x4007, line 1: Will the fixtures be a problem if the UI no longer renders that specific issue?

gentlementlegen: The UI has to render that issue because this is mocked by the "server" (Cypress interceptors). GitHub is actually never called within the tests.

0x4007:
You have some conflicts

cypress/e2e/devpool.cy.ts
```diff
@@ -0,0 +1,146 @@
+describe("DevPool", () => {
+  let issue1: Record<string, unknown>;
+  let issue2: Record<string, unknown>;
+  let loginToken: Record<string, unknown>;
+  let githubUser: Record<string, unknown>;
```
0x4007, line 5: Try setting all the correct types by importing them


Check this example https://github.com/ubiquity/github-labels-processor/blob/5f3921da1fa944d340cd375c3dfc4503531ebee5/src/github-types.ts

cypress/e2e/devpool.cy.ts
```diff
@@ -0,0 +1,146 @@
...
+    cy.get('[for="activity"]').click();
+    cy.get('div[id="issues-container"]').children().should("have.length", 2);
+    cy.get('[for="activity"]').click();
+    cy.get('div[id="issues-container"]').children().should("have.length", 2);
+    cy.get("#filter").type("draft");
+    cy.get('div[id="issues-container"]').children().should("have.length", 2);
+  });
+
+  it("User can log in", () => {
+    cy.intercept("https://api.github.com/repos/*/*/issues**", (req) => {
```
0x4007, line 121: If you are intercepting everything doesn't that kill the point of the test? Why don't we do real end-to-end tests? What if Supabase is broken and we can't oauth etc

cypress/e2e/devpool.cy.ts
```diff
@@ -0,0 +1,146 @@
...
+        statusCode: 200,
+        body: githubUser,
+      });
+    }).as("getUser");
+    cy.intercept("https://iyybhhiflwbsjopsgaow.supabase.co/auth/v1/authorize?provider=github", (req) => {
+      req.reply({
+        statusCode: 200,
+      });
+      // Simulate login token
+      window.localStorage.setItem("sb-wfzpewmlyiozupulbuur-auth-token", JSON.stringify(loginToken));
```
0x4007, line 138: I think real end-to-end testing seems like a missed opportunity here. 

If it's for speed reasons I think it would be cool to have both versions (mocks and real)

Could be interesting depending on how long it takes to run it on lint-staged for any UI code. And the real runs can happen when we push to GitHub

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -1,25 +1,33 @@
...
+      timestamp: Date.now(),
+      tasks: [],
+    };
+  }
+  const cachedTasks = _cachedTasks.tasks.map((task) => ({ ...task, isNew: false, isModified: false })) as TaskMaybeFull[];
   taskManager.syncTasks(cachedTasks);
   if (!cachedTasks.length) {
     // load from network if there are no cached issues
-    return await fetchAndDisplayPreviewsFromNetwork(sorting, options);
+    return fetchAndDisplayPreviewsFromNetwork(sorting, options);
```
0x4007, line 30: Removing await- is that right?

0x4007 (commented):
Cool stuff. Reviewed from my phone.

src/home/fetch-github/fetch-and-display-previews.ts
gentlementlegen: There is nearly no difference between awaiting the return or omitting it. It creates a slightly higher memory use to `return await` because an intermediate Promise is created before the return, and uses one more tick. The only time where it makes a difference is within nested try catch blocks, where scopes of the thrown error will change based on where the await calls are

cypress/e2e/devpool.cy.ts
gentlementlegen: Can do. I didn't bother because they are populated from a string, so type checking is not relevant as I don't manipulate them. Will change it

cypress/e2e/devpool.cy.ts
gentlementlegen: Since we don't develop the other end (eg the backend) I didn't think it was relevant to test it. If GitHub backend breaks there is nothing we can do. The proper way to handle that would be to have BE error test cases, and to have a health check running on the production instance.

cypress/e2e/devpool.cy.ts
gentlementlegen: Same comment as above, it's like testing some code that isn't ours. We could test the speed indeed but it's biased since there is not UI running technically. We could have real e2e testing mixed there, though. The advantage of mocking it is that it is also very easy to setup locally or on a testing machine because no real tokens are needed, and no bandwidth is used.
You can have a read about pros and cons in the interesting piece of documentation
https://docs.cypress.io/guides/guides/network-requests#Use-Server-Responses

src/home/fetch-github/fetch-and-display-previews.ts
0x4007: I see so to clarify: it won't cause issues with resolving the promise?

cypress/e2e/devpool.cy.ts
0x4007: Okay sounds smart I'm in. You can file some issues for this with sufficient details so we can get around to it?

src/home/fetch-github/fetch-and-display-previews.ts
gentlementlegen: Correct! The behavior should be identical.

https://stackoverflow.com/a/42750371/10346405

0x4007:
Just be sure to handle the unresolved comments @FernandVEYRIER

cypress/e2e/devpool.cy.ts
gentlementlegen: You mean an issue mentioning the creation of e2e testing with the real endpoint correct?

0x4007: The proper way as you say

PR #22, opened 2024-03-03
chore: rate limit modal and auto login

### LINKED ISSUE #20: GitHub Rate Limiting Modal on Load
There's a bug that shows the rate limit modal even after logging in. 

### Steps to Reproduce

1. get rate limited as not logged in user. 
2. press the login with github button
3. page automatically reloads with you signed in, but the rate limit modal still appears

at this step we would just refresh and it works fine, but I fear that this may not be intuitive for new users.

### Fix

Please make sure that the page waits to see if you're logged in before displaying the rate limit modal.

Notice: you can get rate limited while logged in so also make sure that this is handled.

Keyrxng:
Resolves #20

- with tasks and no user we display the login rate limit
- With no tasks and no user we display the login rate limit
- with no tasks and a user we force another login (logging in required a fresh to fetch the issues, this prevents that manual refresh)
- with tasks and a user we display an alternative rate limit message
- added a new env var for ``SUPABASE_REF`` and a little helper so it's easier for the likes of myself

I tried to debug why the issues wouldn't load first attempt after logging in following being rate limited but couldn't see why. I presume it's to do with token storage as you clearly see the auth token being handled in the URL. 

I couldn't QA for being limited while auth'd for obv reasons but all bases are covered


0x4007:
> I couldn't QA for being limited while auth'd for obv reasons but all bases are covered

You just need to refresh ~80 times within an hour assuming 30 tasks. I was able to hit these limits while developing the UI. 

Each "task" that is loaded invokes three requests: 1. the preview 2. the full details 3. sometimes, the image (although i have caching for this)

.env.example
```diff
@@ -1,2 +1,3 @@
 SUPABASE_URL=
-SUPABASE_ANON_KEY=
+SUPABASE_REF=
```
0x4007: What is this? There should always be documentation around environment variables, ideally in the README; as this affects setup and the ability for others to work on this. 

Although I don't see why a new env var is necessary in the first place.

cypress/e2e/devpool.cy.ts
```diff
@@ -139,7 +140,7 @@ describe("DevPool", () => {
         statusCode: 200,
       });
       // Simulate login token
-      window.localStorage.setItem("sb-wfzpewmlyiozupulbuur-auth-token", JSON.stringify(loginToken));
+      window.localStorage.setItem(`sb-${getSupabaseRef()}-auth-token`, JSON.stringify(loginToken));
```
0x4007: Is it for testing? Can you rename the property name to make this clear?

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -17,8 +20,23 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
+          // otherwise we have a user and no issues loaded
+          // this happens processing the auth token it seems
+          // as it happens on auth callback and local auth token
+          await gitHubLoginButtonHandler().catch((error) => {
+            console.error(error);
+          });
+        }
+      } else if (user && user !== null) {
+        // Tasks loaded and logged in
+        rateLimitModal(error, `You have been rate limited. Please try again at `);
```
0x4007: I think it makes more sense to directly pass in the full string here, including the timestamp, instead of handling the timestamp generating in the `rateLimitModal` function.

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -34,12 +52,15 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
-function automaticLogin(error: unknown) {
+
+function rateLimitModal(error: unknown, message?: string) {
   const resetTime = error.response.headers["x-ratelimit-reset"];
   const resetParsed = new Date(resetTime * 1000).toLocaleTimeString();
 
   displayPopupMessage(
     `GitHub API rate limit exceeded.`,
-    `You have been rate limited. Please log in to GitHub to increase your GitHub API limits, otherwise you can try again at ${resetParsed}.`
+    !message
```
0x4007: Can you use normal if/else? Ternary is less readable.

0x4007 (changes requested):

.env.example
Keyrxng: I've removed it and just parsing it from the url instead

cypress/e2e/devpool.cy.ts
Keyrxng: It's just for logging in really I did a find/replace but I guess it's useful for testing too, saves having to replace it in the code then replace it again before pushing

src/home/fetch-github/fetch-issues-preview.ts
Keyrxng: removed it completely with just passing in the string and reset time

Keyrxng:
Me trying to get rate limited spamming refresh and calling ``fetchAndDisplayPreviewsFromCache() & fetchIssuesFull()`` x30 lmao

image

I'm seeing no success so I check my rate limit remaining in a separate script, still 5K. 

image

Running it down for QA sake but I'm amazed you were able to hit that locally, I had cache logic disabled and was clearing browser cache every run or two

0x4007:
It must have been from before I implemented cache then. If it's really not going down due to cache then I'd say don't worry about it anymore.

Keyrxng:
Found two scenarios

1. The auth provider can't be reached and returns an error instead of the auth token into the url
2. You get the ``API rate limit exceeded for user ID 106.... If you reach out to GitHub Support for help, ...``

Got the first one maxing it the first time, then number two the second time.

I'd say it's very unlikely to happen but I've added a notice for them anyway.

The ``getNewSessionToken()`` error handling isn't very robust but I wasn't sure whether to log any error into the modal or be explicit with them ``Error getting user profile from external provider`` could be a number of things I suppose

gentlementlegen:
The rate limited cases can be very easily simulated within Cypress by intercepting routes and sending back the appropriate response. It should help you develop and test these cases without having to really spam the endpoint. Also, would be nice to see these testing cases implemented.

Keyrxng:
> The rate limited cases can be very easily simulated within Cypress by intercepting routes and sending back the appropriate response. It should help you develop and test these cases without having to really spam the endpoint. Also, would be nice to see these testing cases implemented.

Makes sense. I'm not as exp with testing as you seem to be and my default is usually: mocking requires pre-knowledge of what's going to happen for a fact, as opposed to simulating a response I never knew existed and even if I do know it exists, mocking it sort of defeats the purpose lmao probs a naïve outlook and why I'm not so great with writing them. Always happy to learn and improve but so 

As part of this PR would you like me to add in test cases to cover what I've done here? @FernandVEYRIER

gentlementlegen:
> > The rate limited cases can be very easily simulated within Cypress by intercepting routes and sending back the appropriate response. It should help you develop and test these cases without having to really spam the endpoint. Also, would be nice to see these testing cases implemented.
> 
> Makes sense. I'm not as exp with testing as you seem to be and my default is usually: mocking requires pre-knowledge of what's going to happen for a fact, as opposed to simulating a response I never knew existed and even if I do know it exists, mocking it sort of defeats the purpose lmao probs a naïve outlook and why I'm not so great with writing them. Always happy to learn and improve but so
> 
> As part of this PR would you like me to add in test cases to cover what I've done here? @FernandVEYRIER

There are JSON example responses in the docs, they don't include failed cases but for each route you have the return codes, which I believe should suffice. That would be really great to have some test for sure!

Keyrxng:
Not a problem, will get on this today

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -66,6 +66,10 @@ export async function fetchAvatars() {
   const urlPattern = /https:\/\/github\.com\/(?<org>[^/]+)\/(?<repo>[^/]+)\/issues\/(?<issue_number>\d+)/;
 
   const avatarPromises = cachedTasks.map(async (task) => {
+    if (!task.preview.body) {
+      return Promise.resolve();
+    }
```
pavlovcik, line 71: What exactly are the consequences of this?

pavlovcik, line 71: What exactly are the consequences of this? It's a bad idea to make unnecessary changes in the pulls.

pavlovcik, line 71: What exactly are the consequences of this? It's a bad idea to make unnecessary changes in the pulls.

cypress/e2e/devpool.cy.ts
```diff
@@ -80,14 +80,69 @@ describe("DevPool", () => {
...
+
+      cy.visit("/");
+      cy.get(".preview-header").should("exist");
+      cy.get(".preview-body-inner").should("include.text", "You have been rate limited. Please try again at");
+    });
+
+    it("Should display a hard one-hour retry timeframe with no auth token available", () => {
+      const urlParams = `#error=server_error&error_code=500&error_description=Error getting user profile from external provider`;
+
+      cy.on("uncaught:exception", () => {
```
gentlementlegen: Can't this be wrapped within an `expect toThrow` check?

src/home/fetch-github/fetch-and-display-previews.ts
pavlovcik, line 71: What exactly are the consequences of this? It's a bad idea to make unnecessary changes in the pulls.

0x4007: What exactly are the consequences of this? It's a bad idea to make out of scope changes in the pulls.

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -79,12 +81,19 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
     if (403 === error.status) {
       console.error(`GitHub API rate limit exceeded.`);
+      const resetTime = error.response.headers["x-ratelimit-reset"];
+      const resetParsed = new Date(resetTime * 1000).toLocaleTimeString();
+
       if (taskManager.getTasks().length == 0) {
-        // automatically login if there are no issues loaded
-        automaticLogin(error);
+        if (!user || user === null) {
+          // only show rate limit modal if there are no issues loaded and not logged in
```
0x4007: I think we should still show the modal even if only some of the issues are loaded.

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +27,38 @@ async function getNewSessionToken(): Promise<string | null> {
...
+      if (code === "500") {
+        if (desc === "Error getting user profile from external provider") {
+          // without a token we can't get a dynamic retry timeframe
+          displayPopupMessage(`GitHub Login Provider`, `Your access token may have reached it's rate limit, please try again after one hour.`);
+          throw new Error("GitHub login provider");
+        }
+      }
+    }
   }
   return providerToken;
```
0x4007: This defaults to null or undefined? I suspect undefined.

src/home/sorting/sorting-manager.ts
```diff
@@ -33,9 +33,12 @@ export class SortingManager {
       const issues = Array.from(issuesContainer.children) as HTMLDivElement[];
       issues.forEach((issue) => {
         const issuePreviewId = issue.children[0].getAttribute("data-preview-id");
-        if (!issuePreviewId) throw new Error(`No preview id found for issue ${issue}`);
+        if (!issuePreviewId) console.error(`No preview id found for issue ${issue}`);
```
0x4007: This shouldn't happen, nor should this be permissible with a console.error. There's a bug that needs to be diagnosed and fixed if this is happening.

src/home/sorting/sorting-manager.ts
```diff
@@ -33,9 +33,12 @@ export class SortingManager {
...
         const fullIssue = taskManager.getTaskByPreviewId(Number(issuePreviewId)).full;
-        if (!fullIssue) throw new Error(`No full issue found for preview id ${issuePreviewId}`);
+        if (!fullIssue) {
+          console.error(`No full issue found for preview id ${issuePreviewId}`);
```
0x4007: Same here.

gentlementlegen (changes requested):

src/home/sorting/sorting-manager.ts
Keyrxng:
```
 1) DevPool
       Items can be sorted:
     Error: The following error originated from your application code, not from Cypress.

  > No task found for preview id 2151943906
```

In the failed cypress run, the id is coming from `fixtures/issue2.json` which links to [this](https://github.com/ubiquity/devpool-directory/issues/1084) repo. I'm unable to reproduce the error now locally throwing the error instead of logging it although this was to prevent the test from failing which seemed to work locally but no effect for the workflow. Perhaps it's related to the GitHub issues will know in a min after this workflow

src/home/getters/get-github-user.ts
Keyrxng: > (method) URLSearchParams.get(name: string): string | null

but I've explicitly returned `token || null`

src/home/fetch-github/fetch-and-display-previews.ts
Keyrxng: `'task.preview.body' is possibly 'null' or 'undefined'.`

I thought that rather than optional chaining to avoid the error/warning then allowing for the rest of the logic to not pass then resolve the promise that it made more sense to resolve if the body was `null` | `undefined` beforehand.

`fetchAvatars()` is only ever returned after having called `displayGitHubIssues()` first which has 

```ts
if (!preview.body) {
  throw new Error(`Preview body is undefined for task with id: ${preview.id}`);
}
```

body being optional comes from the type not the logic as `displayGitHubIssues()` calls `renderGitHubIssues()` which has `if (!existingIssueIds.has(task.preview.id.toString()))` which then calls `everyNewIssue()` which has `if (!taskPreview.preview.body) {return}` and sets `data-preview-id` meaning every issue will have a body and an ID (relating to your other comment) coming from cachedTasks (I thought). 

The cypress run seems to repeatedly be failing for that specific test case however mine are passing, I'm unsure exactly what's causing the other test to fail in the workflow but it's passing locally

cypress/e2e/devpool.cy.ts
Keyrxng: I don't think so as far as I've read from the cypress docs, s/o, gpt and blogs, if I'm mistaken would you mind showing me an example? The only way I've been able to have the test pass is catching the error like that the GUI says that you can disable that behaviour but I didn't think that was the way to go.

I could validate that the thrown error matches the error text but the toaster text is validation of that.

src/home/sorting/sorting-manager.ts
0x4007: @FernandVEYRIER needs to fix the test then. 
- To clarify, theres "mirrored" issues i.e. https://github.com/ubiquity/devpool-directory/issues/1140
- And as you can see they only contain a URL to the "real" issue in its body.
- So we need to parse that URL from the mirror, and then take the full results. 
- "preview" refers to mirrored issues.

gentlementlegen: I believe the mock just needs to be updated with the proper ID. Everything there is mocked so everything is controlled and never reaches the real endpoints. Should be a very minor change fitting in this pull request.

cypress/e2e/devpool.cy.ts
gentlementlegen: I can give a look later. But if that is the only way, make sure to reset this so we don't silence all the errors from the tests, only this specific one.

src/home/sorting/sorting-manager.ts
Keyrxng: resolved in https://github.com/ubiquity/work.ubq.fi/pull/35

cypress/e2e/devpool.cy.ts
gentlementlegen: I think this should be removed, as well as the `throw` inside the function `getNewSessionToken`. We should not make the page crash only because we could not retrieve the token, a `console.error` should suffice.

Keyrxng: removed both mate, the only test failing is the login one which you have a pull open for I'm sure, yeah?

gentlementlegen: I believe it would be this one https://github.com/ubiquity/work.ubq.fi/pull/35
Let's merge it first so you can rebase and make sure everything works as expected.

Keyrxng: same issue still but it's expected to happen from time to time is that right?

```
   1) DevPool
       User can log in:
     CypressError: `cy.type()` can only accept a string or number. You passed in: `undefined`
```

nvm it was an env issue realise that now, should be all good

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +27,38 @@ async function getNewSessionToken(): Promise<string | null> {
...
   const providerToken = params.get("provider_token");
   if (!providerToken) {
-    return null;
+    const err = params.get("error");
+    const code = params.get("error_code");
+    const desc = params.get("error_description");
+
+    if (err === "server_error") {
+      if (code === "500") {
+        if (desc === "Error getting user profile from external provider") {
```
gentlementlegen: Is there no better check that this? If one letter in the text changes it breaks the whole login. Also `500` and `server_error` seem pretty redundant.

gentlementlegen (changes requested):

src/home/getters/get-github-user.ts
Keyrxng: I don't think so no but I've removed the err and code checks

It's the Supabase response so not much control over it, the test-case I found it in was that I had maxed out PAT

gentlementlegen: I mean that the only check here should be `code === '500'`. `err` is just the translation of the `code` to readable text, and `desc` just sent by Supabase as a default error message.

You can use it to output the error instead:
`console.error(desc);`

Keyrxng: Okay mate I understand, I was trying to be specific to the edge-case that way I wasn't handling other failures with the token expired response. Because ``code == 500`` could represent multiple errors, I remember mentioning it previously re: generic vs specific approach to handling it.

But what'll do is just `console.err` the desc if `code == 500` and fire the toaster regardless

Keyrxng: checking only that `code === '500' and if so console.err and show toaster to user

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +27,33 @@ async function getNewSessionToken(): Promise<string | null> {
   const params = new URLSearchParams(hash.substr(1)); // remove the '#' and parse
   const providerToken = params.get("provider_token");
   if (!providerToken) {
-    return null;
+    const code = params.get("error_code");
+
+    // supabase auth provider has failed for some reason
+    if (code === "500") {
+      displayPopupMessage(`GitHub Login Provider`, `Your access token may have reached it's rate limit, please try again after one hour.`);
```
gentlementlegen:
```suggestion
      displayPopupMessage(`GitHub Login Provider`, `Your access token may have reached its rate limit, please try again after one hour.`);
```

gentlementlegen (changes requested):

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -79,12 +80,16 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
   } catch (error) {
     if (403 === error.status) {
       console.error(`GitHub API rate limit exceeded.`);
-      if (taskManager.getTasks().length == 0) {
-        // automatically login if there are no issues loaded
-        automaticLogin(error);
+      const resetTime = error.response.headers["x-ratelimit-reset"];
+      const resetParsed = new Date(resetTime * 1000).toLocaleTimeString();
+
+      if (!user || user === null) {
```
gentlementlegen: Shouldn't `(!user)` be enough for this check?

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +27,33 @@ async function getNewSessionToken(): Promise<string | null> {
...
-async function getNewGitHubUser(providerToken: string): Promise<GitHubUser> {
+async function getNewGitHubUser(providerToken: string | null): Promise<GitHubUser | null> {
   const octokit = new Octokit({ auth: providerToken });
-  const response = (await octokit.request("GET /user")) as GitHubUserResponse;
-  return response.data;
+  try {
+    const response = (await octokit.request("GET /user")) as GitHubUserResponse;
+    return response.data;
+  } catch (error: unknown) {
+    if (error.status === 403 && error.message.includes("API rate limit exceeded")) {
```
gentlementlegen: Instead of checking the error message string, I think it might be wiser to call https://octokit.github.io/rest.js/v20#rate-limit which seems to be there for this purpose?

If it works fine, the same principle can be applied on line 34.

src/home/rendering/render-github-login-button.ts
```diff
@@ -18,7 +18,7 @@ export async function checkSupabaseSession() {
   return session;
 }
 
-async function gitHubLoginButtonHandler() {
+export async function gitHubLoginButtonHandler() {
```
gentlementlegen: Any reason for this export? I do not see it called anywhere in the content of this pull request.

gentlementlegen (changes requested):

src/home/getters/get-github-user.ts
Keyrxng: This only works for the line you've suggested here as in line 34 we do not have an `auth-token` as the `auth provider` (supabase gh integration) has returned an error string instead of the user's token

I can use that endpoint in both but default to a hardcoded `1 hour` if we don't have a token in L34

src/home/rendering/render-github-login-button.ts
Keyrxng: Before the auth setup was changed from relying on `localStorage` I was using this to force a login to remove the burden from the user in having to manually refresh, as it's no longer needed I'll remove the export

src/home/getters/get-github-user.ts
Keyrxng: realized that we have access to the header response regardless of `auth-token` and the call flow allows us to use those headers, so I've removed the hard `1 hour` outcome

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -97,12 +92,44 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
+      rate.user = (await getGitHubUser()) ? true : false;
+    } catch (err) {
+      console.error("Error handling GitHub rate limit", err);
+    }
+  }
+
+  const resetParsed = rate.reset && new Date(rate.reset * 1000).toLocaleTimeString();
+
+  if (!rate.user) {
+    rateLimitModal(`You have been rate limited. Please log in to GitHub to increase your GitHub API limits, otherwise you can try again at ${resetParsed}.`);
```
gentlementlegen, line 132: Does this take the UTC into account when displayed?

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +27,25 @@ async function getNewSessionToken(): Promise<string | null> {
...
+async function getNewGitHubUser(providerToken: string | null): Promise<GitHubUser | null> {
   const octokit = new Octokit({ auth: providerToken });
-  const response = (await octokit.request("GET /user")) as GitHubUserResponse;
-  return response.data;
+
+  try {
+    const response = (await octokit.request("GET /user")) as GitHubUserResponse;
+    return response.data;
+  } catch (error: unknown) {
+    if (error.status === 403) {
```
gentlementlegen: This displays a TypeScript error as `error` is `unknown`

gentlementlegen (changes requested):

src/home/fetch-github/fetch-issues-preview.ts
0x4007: > .toLocaleTimeString();

It uses local time. This is relevant for the user, and in my testing, it is accurate.

Just realized, this pull needs to be rebased/reopened. Pretty sure this is my code from a long time ago that's already in production.

Keyrxng: - https://github.com/ubiquity/work.ubq.fi/blame/ecb2187ddc2067e99ae806af3430c67bf19a8d9d/src/home/fetch-github/fetch-issues-preview.ts#L100

The code is similar I guess but this pr is up to date with development if that's where prod is running from

src/home/getters/get-github-user.ts
Keyrxng: I stopped ts from typing catch() as `unknown` by default allowing the use of `any`.

`unknown` is safer but if we are going to have to cast it anyway to stop the error and we know what we are receiving then `any` makes as much sense to use

gentlementlegen: While I am not against the idea, I think it is better to check the type with `instanceof` and use the Octokit provided [error library](https://github.com/octokit/request-error.js?tab=readme-ov-file#usage-with-octokit) to avoid recreating an error within the catch blocks

gentlementlegen (changes requested):

src/home/getters/get-github-user.ts
Keyrxng: implemented

gentlementlegen (approved):
Good to go when @pavlovcik is

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +28,22 @@ async function getNewSessionToken(): Promise<string | null> {
...
+  return providerToken || null;
 }
 
-async function getNewGitHubUser(providerToken: string): Promise<GitHubUser> {
+async function getNewGitHubUser(providerToken: string | null): Promise<GitHubUser | null> {
   const octokit = new Octokit({ auth: providerToken });
-  const response = (await octokit.request("GET /user")) as GitHubUserResponse;
-  return response.data;
+  try {
+    const response = (await octokit.request("GET /user")) as GitHubUserResponse;
```
0x4007, line 41: Is this the best method to use to get the user? Doesn't octokit have a dedicated method?

src/home/getters/get-github-user.ts
```diff
@@ -30,13 +28,22 @@ async function getNewSessionToken(): Promise<string | null> {
...
+    return response.data;
+  } catch (error) {
+    if (error instanceof RequestError && error.status === 403) {
+      await handleRateLimit(providerToken ? octokit : undefined, error);
```
0x4007, line 45:
```suggestion
      await handleRateLimit(providerToken ?? octokit, error);
```
Is this more concise?

src/home/getters/get-github-user.ts
gentlementlegen: It does but afaik it only works with username, not the user id
https://octokit.github.io/rest.js/v20#users-get-by-username
Using request seems to be the only way

src/home/getters/get-github-user.ts
Keyrxng: Although if there is a `providerToken` and we get an error it'll pass `providerToken` as an arg instead of `octokit` which would mean rebuilding `octokit` within `handleRateLimit()` while both invocations of `handleRateLimit()` already have `octokit` built, one is guaranteed to have a provider token and this one in question is not.

If we make this more concise we litter the rate limit function

I guess `if(typeof octokit == "string")` would work instead of `if(octokit){}else{}` but typing octokit as string seemed daft when it first popped into my head 😂

Keyrxng:
Can this be merged? @0x4007 @gentlementlegen

PR #23, opened 2024-03-03
chore: fetch personal pp and remove notifications

### LINKED ISSUE #21: March Adjustments
image

1. No new issue indicators
2. Personal GitHub image doesn't show (it's built for organizations only)

Keyrxng:
Resolves #21 

- failing to fetch the org pp will fetch from the users endpoint
- added ``NOTIFICATION_EXPIRY_DAYS = 3`` 
- it seemed that previously the ``isNew`` prop was being wiped after it experienced it's first cache refresh, this way it's based on the ``created_at`` which seems like a better approach

The timeframe looks a little squished when it's at defcon 1 don't you think?

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,4 +36,3 @@ open http://localhost:8080
 #### Mobile
 
 ![screenshot 2](https://github.com/ubiquity/devpool-directory-ui/assets/4975670/b7861ce7-1f1f-49a9-b8e2-ebb20724ee67)
-
--- a/src/home/fetch-github/fetch-and-display-previews.ts
+++ b/src/home/fetch-github/fetch-and-display-previews.ts
@@ -14,6 +14,8 @@ export type Options = {
   ordering: "normal" | "reverse";
 };
 
+const NOTIFICATION_EXPIRY_DAYS = 3;
+
 export async function fetchAndDisplayPreviewsFromCache(sorting?: Sorting, options = { ordering: "normal" }) {
   let _cachedTasks = getLocalStore(GITHUB_TASKS_STORAGE_KEY) as TaskStorageItems;
   // makes sure tasks have a timestamp to know how old the cache is, or refresh if older than 15 minutes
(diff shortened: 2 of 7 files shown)


0x4007:
Let's make it simpler. Can we just remove the new issue indicator? The CSS and TS code for it. 

I think in order to properly implement it, the caching needs to be rebuilt entirely, because I poorly implemented the caching the first time around. 

> The timeframe looks a little squished when it's at defcon 1 don't you think?

I was thinking the same. Do `text-overflow: ellipsis; overflow: hidden; display: inline-block;` on all labels. Test on mobile and desktop to make sure that `inline-block` doesn't screw anything up.

> it seemed that previously the isNew prop was being wiped after it experienced it's first cache refresh, this way it's based on the created_at which seems like a better approach

No the intent is to ensure that its new for the specific user viewing on the client. Not if its a new issue for the network. Its supposed to act like a social media notification in a sense. Or like a GitHub notification.

Keyrxng:
I understand the notifications now, the vid shows them working as described. I delete them from my local storage to show it twice. Would you rather I just remove it all completely?


https://github.com/ubiquity/work.ubq.fi/assets/106303466/c4391a87-9e04-4582-87a9-41927f4a1df5

mobile_inline_block


desktop_inline_block

0x4007 (approved):

0x4007:
I just checked again on my computer and see that CI failed. Just some minor cleanup please

https://github.com/ubiquity/work.ubq.fi/actions/runs/8162423256/job/22318480984#step:4:93

0x4007:
Strange, the error did not appear in the latest commit deployment for me when I tried in my browser. @FernandVEYRIER any ideas why this would be? I tried incognito as well.

gentlementlegen:
Not sure, can have a look if that error persist. You can always check the artifacts as well within the Action, like here
https://github.com/ubiquity/work.ubq.fi/actions/runs/8162423256
According to the screenshot the div indeed does not have the `new-task` styling, somehow

0x4007:
Yes I noticed the screenshot as well. Was curious if there's benefit to doing videos as well, as I saw that as an option in the config.

0x4007:
Is it possible that Cypress is pulling in the wrong commit? This seems the most likely. However the first deployment doesn't even seem stable.

gentlementlegen:
Sure I can enable the video, I didn't because it uses a lot of storage and I didn't know about the current limitations. I don't think Cypress would pull the wrong commit, will look into it

0x4007:
> I didn't know about the current limitations

Since we're open source, GitHub sort of gives us unlimited storage etc.

0x4007:
@FernandVEYRIER I just realized you wrote custom tests for this UI. So this makes sense. You specifically requested that it checks for `new-task` which was removed here.

PR #25, opened 2024-03-06
Fix/cypress test

0x4007:
- Fixed the broken test from https://github.com/ubiquity/work.ubq.fi/pull/23#issuecomment-1979854728
- Added types so VSCode doesn't complain in the test file.


gentlementlegen:
@pavlovcik Thanks for your help. I was curious, but why Jest package was added here, there is no Jest test is it? Also, the new items now have the exact same visual as the old items that were previously added in the item list?

0x4007:
> @pavlovcik Thanks for your help. I was curious, but why Jest package was added here, there is no Jest test is it?

I assumed the types were jest in the cypress tests. That's why I confirmed with a pr and requested you to review. Perhaps you can take over and merge this to confirm the best practices/settings. But vscode was complaining about types prior to this pull. 

> Also, the new items now have the exact same visual as the old items that were previously added in the item list?

I don't understand but the is-new stuff was removed in the other pull today.

gentlementlegen:
> > @pavlovcik Thanks for your help. I was curious, but why Jest package was added here, there is no Jest test is it?
> 
> I assumed the types were jest in the cypress tests. That's why I confirmed with a pr and requested you to review. Perhaps you can take over and merge this to confirm the best practices/settings. But vscode was complaining about types prior to this pull.
> 
> > Also, the new items now have the exact same visual as the old items that were previously added in the item list?
> 
> I don't understand but the is-new stuff was removed in the other pull today.

Cypress uses Mocha and Chai for its testing, you can have a look [here](https://docs.cypress.io/guides/references/bundled-libraries). I am not using VSCode myself, but will have a look for the types then. Webstorm recognizes them out of the box.

Before when new issues (and so new items) were created, there was a blue styling thing on the items, which is what I was checking for, so indeed if this was removed that is why the test failed, thanks for fixing.

gentlementlegen:
@pavlovcik I opened the project with VS Code, removed Jest types and have no trouble with the auto-complete, cf screenshot
image
what kind of errors are you experiencing? I cannot seem to be able to repro on my side

0x4007:
It was only red squiggles. Let's add the minimum settings (possibly just tsconfig types) and merge this in

gentlementlegen:
@pavlovcik You got it. Opened a PR with the fixes for this one, at https://github.com/pavlovcik/work.ubq.fi/pull/1 and repaired some tests in the meantime.

gentlementlegen:
I am not sure if the tests are broken or if it's again GitHub displaying the previously errored tests. It works pretty fine for me locally

gentlementlegen (approved):

PR #27, opened 2024-03-14
feat: text field is now hidden when not logged in

### LINKED ISSUE #26: Only Render Text Filter After Logged In
The text filter in the toolbar doesn't seem to work very well until the full task details are loaded (usually requires being logged in due to increased limits.)

Perhaps it makes sense to hide it until they are logged in. That way it will be easier to see the login with GitHub button on mobile!

work ubq fi_ (2)

gentlementlegen:
Resolves #26

### DIFF
--- a/static/style/style.css
+++ b/static/style/style.css
@@ -579,6 +579,9 @@
     font-size: 12px;
     letter-spacing: 0.5px;
   }
+  #filters input[type="text"].hidden {
+    visibility: hidden;
+  }
   button#github-login-button {
     margin-left: 0;
   }
--- a/static/style/style.css
+++ b/static/style/style.css
@@ -580,7 +580,7 @@
     letter-spacing: 0.5px;
   }
   #filters input[type="text"].hidden {
-    visibility: hidden;
+    display: none;
   }
   button#github-login-button {
     margin-left: 0;
--- a/.github/workflows/cypress-testing.yml
+++ b/.github/workflows/cypress-testing.yml
@@ -2,7 +2,7 @@ name: Run Cypress testing suite
 on:
   workflow_dispatch:
   pull_request:
-    types: [ opened, synchronize ]
+    types: [opened, synchronize]
 
 jobs:
   cypress-run:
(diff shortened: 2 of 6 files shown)


gentlementlegen:
What are the changes of this PR:
- text is hidden while logged out, visible when logged in
- the tests were all broken due to the use of OAuth with Supabase, before we use to simply put the token in the localStorage, now the session is really retrieved from Supabase `auth`
- while this is a clear improvement, it broke the tests as in I just faked the local storage before. Now, there is a real OAuth workflow going on. Sadly, Oauth cannot be mocked.
- dummy account was created for this purpose, credentials given to @pavlovcik and @rndquu, should be added to the env `UBIQUIBOT_GITHUB_USERNAME` and `UBIQUIBOT_GITHUB_PASSWORD`

static/style/style.css
```diff
@@ -579,6 +579,9 @@
     font-size: 12px;
     letter-spacing: 0.5px;
   }
+  #filters input[type="text"].hidden {
+    visibility: hidden;
```
0x4007:
```suggestion
    display: none;
```

0x4007: I did this because it will not create an empty space in the DOM, which seems to make more sense for this purpose.

0x4007:
I know that @rndquu changed the Cloudflare deploys. You have the final say on whether to approve and merge this pull!

rndquu (approved):

rndquu:
Works fine

> I know that @rndquu changed the Cloudflare deploys.

I haven't yet updated all of the repos with updated build workflow but current PR works fine and can be merged

0x4007:
Seems like it's not on production I tried refreshing several times

gentlementlegen:
> Seems like it's not on production I tried refreshing several times

Still the case so far, got news about this?

0x4007:
@rndquu might know best

rndquu:
> > Seems like it's not on production I tried refreshing several times
> 
> Still the case so far, got news about this?

Working on it

PR #36, opened 2024-03-20
Chore/background gradient update

0x4007:
Matches pay.ubq.fi styles


0x4007:
Help button doesn't look too great by itself in the footer but the rest of the page looks pretty solid.

PR #39, opened 2024-03-23
feat: track devRel referrals with google analytics

### LINKED ISSUE #30: GitHub Registration DevRel Tracker
### Registration Tracker

- As we ramp up our recruiting efforts, we will provide incentives for developer relations managers to spread the word of the DevPool.
- A major incentive is to reward the DevRel responsible for onboarding any developers that we end up hiring. 
- I anticipate that they can coax and encourage the new candidate to finish one task to make it to getting paid. We may even be able to automate payment generation for the DevRel in the future!

### Scope

We should enable support for trackers i.e. `work.ubq.fi?devrel=diamondnegroni` and then when the developer signs up, we capture the dev and associate them with the devrel for future reference. Refer to https://github.com/ubiquity/work.ubq.fi/issues/30#issuecomment-2008148491 for implementation details. 

# Deadline = EthSeoul Conference 25 March 2024!

EresDev:
Resolves #30

This should be merged ASAP so that we can test google analytics receiving events. Because my test domain was different than the production domain. 

The link is going to be something like `https://work.ubq.fi/?devRel=XYZ` where `XYZ` is the code you want to give to DevRel. This code can't be the word `done` because it has a special meaning in code. 

There are some things I want to refactor like: 
- moving the measurement id of Google Analytics to configs
- getting rid of `@ts-expect-error` on the use of gtag()

But I am very tired and out of energy right now. Google Analytics gave me a tough time and I am still not fully satisfied with its results. 

Right now the foucs should be on testing it. Making it work in the conference. Any refactoring you want can be done in the next PR, leave comments for that.

QA:

I am hoping to get something like this out of Google Analytics: 

image

This is from yesterday's data. It is bad and it is what I could get, and I was hoping to get a better report of today's data but I will access to that data tomorrow in Google Analytics Explore. 

The only thing that gives me a little confidence is the event being received by Google Analytics with correct data. We can hope to give it a better shape. 

Screenshot_20240323_221538

### DIFF
--- a/src/home/authentication.ts
+++ b/src/home/authentication.ts
@@ -1,3 +1,4 @@
+import { trackDevRelReferral } from "./devrel-tracker";
 import { getGitHubAccessToken } from "./getters/get-github-access-token";
 import { getGitHubUser } from "./getters/get-github-user";
 import { GitHubUser } from "./github-types";
@@ -12,6 +13,7 @@ export async function authentication() {
 
   const gitHubUser: null | GitHubUser = await getGitHubUser();
   if (gitHubUser) {
+    trackDevRelReferral(gitHubUser.name as string);
     displayGitHubUserInformation(gitHubUser);
   }
 }
--- a/src/home/devrel-tracker.ts
+++ b/src/home/devrel-tracker.ts
@@ -1,0 +1,22 @@
+export function initiateDevRelTracking() {
+  const oldDevRelCode = localStorage.getItem("devRel");
+  if (!oldDevRelCode) {
+    const urlParams = new URLSearchParams(window.location.search);
+    const devRelCode = urlParams.get("devRel");
+    if (devRelCode) {
+      localStorage.setItem("devRel", devRelCode);
+    }
+  }
+}
+
+export function trackDevRelReferral(devGithub: string) {
+  const devRelCode = localStorage.getItem("devRel");
+  if (devRelCode && devRelCode != "done") {
+    // @ts-expect-error : using global gtag
+    gtag("event", "ethSeoul_registration", {
+      devRel: devRelCode,
+      devGithub: devGithub,
+    });
+    localStorage.setItem("devRel", "done");
+  }
+}
--- a/src/home/home.ts
+++ b/src/home/home.ts
@@ -1,11 +1,13 @@
 import { grid } from "../the-grid";
 import { authentication } from "./authentication";
+import { initiateDevRelTracking } from "./devrel-tracker";
 import { fetchAndDisplayPreviewsFromCache } from "./fetch-github/fetch-and-display-previews";
 import { fetchIssuesFull } from "./fetch-github/fetch-issues-full";
 import { readyToolbar } from "./ready-toolbar";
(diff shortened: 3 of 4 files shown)


EresDev:
Now I can access yesterday's data, and I see some better Explore reports. 

image

image


There was a chance of duplication as you see above EresDev signed up twice with the same link. There probably would be a way to handle that in Google Analytics but to be on the safe side, I covered that in code. 

If we can merge soon, we probably will see some more data.

rndquu (approved):

EresDev:
I just tested it and it is working on live. One thing that I Just noticed is that in the URL `https://work.ubq.fi/?devRel=XYZ` the query parameter devRel is case-sensitive. The `R` in devRel must be capitalized. It would have been nice to have it case-insensitive but this is what is working now. Just be careful.  @pavlovcik

rndquu:
@EresDev So we should set production GA token [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR8) and [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR16), right? Or `G-VZLJ61H1YM` is already the production one used by Ubiquity organization?

EresDev:
> @EresDev So we should set production GA token [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR8) and [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR16), right? Or `G-VZLJ61H1YM` is already the production one used by Ubiquity organization?

This is correct if you want to change it. This is a code from my account and not Ubiquity organization account. But @pavlovcik wanted me to export the reports and I will make him an admin too on the Google Analytics project. So it is preferred to keep the current one per our discussion. But if you want to change it, then you can.

rndquu:
> > @EresDev So we should set production GA token [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR8) and [here](https://github.com/ubiquity/work.ubq.fi/pull/39/files#diff-3e8602c86d6cf04a5c74b954f9f957b652617102cf8f4b647fe46989f480861dR16), right? Or `G-VZLJ61H1YM` is already the production one used by Ubiquity organization?
> 
> This is correct if you want to change it. This is a code from my account and not Ubiquity organization account. But @pavlovcik wanted me to export the reports and I will make him an admin too on the Google Analytics project. So it is preferred to keep the current one per our discussion. But if you want to change it, then you can.

So we're going to use your Google Analytics account and then you'll simply export the report, right? In this case there is no need to update anything.

EresDev:
> So we're going to use your Google Analytics account and then you'll simply export the report, right? In this case there is no need to update anything.

Correct.

PR #42, opened 2024-04-11
Reduce Requested Permission for new comers

### LINKED ISSUE #41: Lower requested permissions on user registration
When a new user registers to the website and logs in via GitHub, sensitive rights are requested by the app, which is `read / write` on all `public / private` repositories. This seems to be excessive and worrisome for newcomers as they would wonder why we need to write on their private repos. Ideally, we should just request for read permission on public repos. I do not know myself the reason of the request of such elevated rights, so it would be worth investigating. If elevated rights are indeed needed for core members or insiders of the organization, these rights should be requested specifically for these members not for every new user registering.

image

jordan-ae:
Resolves #41


jordan-ae:
@0x4007 @gentlementlegen This should reduce requested permission for new comers

.vscode/extensions.json
gentlementlegen, line 1: I don't think this should be committed

.vscode/settings.json
gentlementlegen, line 1: I don't see the use to have this file committed either

database.types.ts
gentlementlegen, line 1: Please remove empty files

package-lock.json
gentlementlegen, line 1: The preferred package manager is `yarn`. You might notice there is already a `yarn.lock` file

supabase/config.toml
gentlementlegen, line 1: The supabase folder should not be committed here because it is actually contained in another repo:
https://github.com/ubiquity/ubiquibot
Please pull and use this one to benefit from Supabase CLI.

src/home/rendering/render-github-login-button.ts
```diff
@@ -19,16 +19,48 @@ export async function checkSupabaseSession() {
...
+        scopes: "public_repo", // Default scope for newcomers
+      },
+    });
+
+    if (error) {
+      console.error("Error logging in:", error);
+      return;
+    }
+
+    const { data, error: fetchError } = await supabase.from("users").select("userType").eq("id", user.id).single();
```
gentlementlegen, line 35: We do not have such column for users. However their `id` matches their GitHub ID. From there, you can fetch the user details and check they level inside the Organization, `member` or `collborator` etc.

gentlementlegen (changes requested):

src/home/rendering/render-github-login-button.ts
jordan-ae: Please whats the column users are stored on in the supabase instance? Thanks

jordan-ae: The architecture of the project and how all the pieces communicate is a bit confusing to me. Please how can I get the schema correctly. Thanks

gentlementlegen: The project architecture looks something similar to this
https://github.com/ubiquibot/.github/pull/11#issuecomment-2041143856

And the Supabase schema can be found at
https://github.com/ubiquity/ubiquibot/tree/development/supabase

rndquu:
Closing this one, already fixed by https://github.com/ubiquity/work.ubq.fi/pull/48

PR #48, opened 2024-05-10
fix: lowered user permissions on registration

### LINKED ISSUE #41: Lower requested permissions on user registration

### LINKED ISSUE #47: Empty directory page after logged in for the first time
## Summary
When a newcomer logs in with GitHub authentication, the directory pool appears as an empty page.

## Description
Upon login, the page requests collaborator information for the logged-in user from the repository by calling `/repos/ubiquity/devpool-directory-private/collaborators/{username}`. This request receives a `404 NotFound` result, which is expected if the user is not a collaborator.

**Root Cause Analysis (from `jordan-ae`'s investigation):**
The issue is not related to caching. The error occurs within `fetch-issues-preview.ts`. The function attempts to fetch private issues first. For collaborators, the line `hasPrivateRepoAccess = await checkPrivateRepoAccess();` triggers a `403 error`. Consequently, the application bypasses the check for public repository issues and jumps to the catch clause, executing `handleRateLimit(octokit, error)`. This sets the `task` to `freshIssues`, which is an empty array due to the fetch failure. This empty task is then cached and displayed, resulting in a blank page.

This bug was introduced in PR #20.

## Expected Behavior
Even if the user is not a collaborator, the devpool listing should still be accessible on the listing page, similar to when the user is not logged in. The system should gracefully handle the `403 error` for private repo access and proceed to fetch public repo issues.

## Proposed Solution
The issue has been identified and a fix has been implemented in PR #48 (specifically `https://github.com/ubiquity/work.ubq.fi/pull/48/files#diff-a526a47f6685dd952321fd365cea2c75b28fb920d5d3f46f3dedb61d0ebed95bR27`). The solution involves adjusting the logic in `fetch-issues-preview.ts` to correctly handle the `403 error` for private repository access and ensure public issues are still fetched and displayed.

gentlementlegen:
Resolves #41 
Resolves #47

### DIFF
--- a/src/home/authentication.ts
+++ b/src/home/authentication.ts
@@ -1,14 +1,30 @@
+import { Octokit } from "@octokit/rest";
 import { trackDevRelReferral } from "./devrel-tracker";
 import { getGitHubAccessToken } from "./getters/get-github-access-token";
 import { getGitHubUser } from "./getters/get-github-user";
 import { GitHubUser } from "./github-types";
 import { displayGitHubUserInformation } from "./rendering/display-github-user-information";
-import { renderGitHubLoginButton } from "./rendering/render-github-login-button";
+import { renderAugmentAccessButton, renderGitHubLoginButton } from "./rendering/render-github-login-button";
+
+/**
+ * Checks if the logged-in user is part of Ubiquity's Org, and didn't grant the 'repo' scope
+ */
+async function isOrgMemberWithoutScope(accessToken: string) {
+  const octokit = new Octokit({ auth: accessToken });
+  const { headers } = await octokit.request("HEAD /");
+  if (headers) {
+    const scopes = headers["x-oauth-scopes"]?.split(", ");
+    return !scopes?.includes("repo");
+  }
+  return false;
+}
 
 export async function authentication() {
   const accessToken = await getGitHubAccessToken();
   if (!accessToken) {
     renderGitHubLoginButton();
+  } else if (await isOrgMemberWithoutScope(accessToken)) {
+    renderAugmentAccessButton();
   }
 
   const gitHubUser: null | GitHubUser = await getGitHubUser();
--- a/src/home/authentication.ts
+++ b/src/home/authentication.ts
@@ -1,35 +1,19 @@
-import { Octokit } from "@octokit/rest";
 import { trackDevRelReferral } from "./devrel-tracker";
 import { getGitHubAccessToken } from "./getters/get-github-access-token";
 import { getGitHubUser } from "./getters/get-github-user";
 import { GitHubUser } from "./github-types";
 import { displayGitHubUserInformation } from "./rendering/display-github-user-information";
-import { renderAugmentAccessButton, renderGitHubLoginButton } from "./rendering/render-github-login-button";
-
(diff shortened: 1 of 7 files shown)


gentlementlegen:
This is a proposal to fix the permissions requested for newcomers, as requesting full access on public / private repos seems excessive and scary for new users.

It's a bit complicated to solve, since knowing the data related to a user requires that user to be logged in, where scopes have already been requested. So here is a draft proposal on how to handle it:

1. A new user logs in for the first time at `work.ubq.fi` and gets requested the following
image

Everything is read-only, and the Org will help us later on determine if that user if part of Ubiquity's Org.

2. We check post login if that user is part of Ubiquity's team. If so, we display a "lock" button next to the user name, like so

image


When clicked, we get redirected to the GitHub OAuth permission page requesting higher privileges

image

If allowed, the "lock" button is gone.

If that workflow is satisfactory, I can spend some time on the CSS to make it look better.

0x4007:
I think it's a great proposal thanks for figuring this out. I am unsure that a lock is the best icon but until I have time to look through icon libraries I am unsure which is better. 

Try using Google material design icons I normally stick with those for all of our UIs.

0x4007:
I think you should also set a time estimate.

0x4007:
Oh I just realized this is a pull sorry!

gentlementlegen:
@0x4007 Thanks for the feedback. Seems that Ubiquibot still doesn't like me: https://github.com/ubiquity/work.ubq.fi/issues/41#issuecomment-2104200066

For the lock icon I actually pulled it out from [Material UI icons](https://mui.com/material-ui/material-icons/?query=lock). Maybe an opened lock would make more sense anyway.
By CSS issue I was referring that the placement of the button is inappropriate right now as it takes too much left and right space with the user name.

molecula451 (approved):

gentlementlegen:
Current looks of the button in the navbar:
image

rndquu (approved):

gitcoindev (approved):

gitcoindev:
Looks good! I also reviewed from my side.

0x4007:
> Resolves #41 
> 
> Resolves #47 
> 
> 
> 
> 

We discourage multiple associated issues in a single pull. 

This is so that we can audit changes (git blame) more efficiently if problems are discovered later. 

In the future I would branch off and open two pulls separately. This can be done retroactively after all the code is complete as long as you use the git "cherry pick" feature @gentlementlegen

gentlementlegen:
@0x4007 Yep makes sense, very accidentally fixed two by doing this one. Thought it would not make much sense to have another PR with the same exact code as this one, will keep this in mind and open two different PRs next time.

PR #56, opened 2024-06-03
Add default placeholder when avatar fails to load

### LINKED ISSUE #55: Render Default Image when Avatar Fails to Load
When avatars fail to load, I think it should render a default image or something. Right now when default avatars aren't fetched it doesnt display anything. So the user cant log out because the avatar doubles down as logout button.

Screenshot 2024-06-03 192445

jordan-ae:
Resolves #55

### DIFF
--- a/src/home/rendering/display-github-user-information.ts
+++ b/src/home/rendering/display-github-user-information.ts
@@ -14,9 +14,7 @@ export async function displayGitHubUserInformation(gitHubUser: GitHubUser) {
   if (gitHubUser.avatar_url) {
     img.src = gitHubUser.avatar_url;
   } else {
-    img.style.width = "50px";
-    img.style.height = "50px";
-    img.style.backgroundColor = "#808080";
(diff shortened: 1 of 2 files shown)


src/home/rendering/display-github-user-information.ts
```diff
@@ -11,7 +11,13 @@ export async function displayGitHubUserInformation(gitHubUser: GitHubUser) {
   if (!toolbar) throw new Error("toolbar not found");
 
   const img = document.createElement("img");
-  img.src = gitHubUser.avatar_url;
+  if (gitHubUser.avatar_url) {
+    img.src = gitHubUser.avatar_url;
+  } else {
+    img.style.width = "50px";
```
0x4007: Style is to be handled by CSS. Also are you sure it renders at 50px? I would assume 48px.

0x4007 (changes requested):

0x4007 (approved):

PR #60, opened 2024-06-04
fix: cypress now runs on pull_request

### LINKED ISSUE #58: Fix Cypress Tests
> The last successful run of Cypress was [three months ago](https://github.com/ubiquity/work.ubq.fi/actions/runs/8199248704) with a single test, so it appears as though they were never stable in the first place @gentlementlegen 

> Perhaps the Cypress tests should all be fixed as part of a separate task.

_Originally posted by @0x4007 in https://github.com/ubiquity/work.ubq.fi/issues/57#issuecomment-2147352850_

gentlementlegen:
Resolves #58


gentlementlegen:
## What is this?

Unlike when we are using Jest, Cypress cannot stub Supabase for auth, because we use a bundler that will create its own code, which is not accessible through Cypress itself. As such, it is not possible to fake login, or auth sessions.

Every tutorial I found suggests to have a separate Supabase session instantiated within Cypress, and store the credentials within the browser. This introduces 2 problems:
- we would have to enable credentials login, since we only have OAuth
- we would have to store valid credentials, which we can't since we do not have access to secrets

The workaround I implemented for this is to read the `NODE_ENV` value and if set to `test` then we would just read the `token` from the local storage, because we are able to manipulate this through Cypress thus fake login.

For the rest just fixed the tests, added missing interceptions, and fixed toast conditions. Might collide with https://github.com/ubiquity/work.ubq.fi/pull/57
Successful run: https://github.com/ubiquity/work.ubq.fi/actions/runs/9371378749

0x4007:
> Might collide with https://github.com/ubiquity/work.ubq.fi/pull/57
> 
> Successful run: https://github.com/ubiquity/work.ubq.fi/actions/runs/9371378749

Given that #57 was opened first and is the reason this task was created, it makes sense for you to accept it first and then accommodate necessary changes here

rndquu (approved):

0x4007:
> > Might collide with #57
> > Successful run: https://github.com/ubiquity/work.ubq.fi/actions/runs/9371378749
> 
> Given that #57 was opened first and is the reason this task was created, it makes sense for you to accept it first and then accommodate necessary changes here

Might be appropriate to properly configure a merge queue. Although I couldn't figure it out last time i tried @gentlementlegen

whilefoo (approved):
I'm not too familiar with this codebase but it looks good to me

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -65,16 +64,10 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
+      await handleRateLimit(octokit, error as RequestError);
     } else {
-      // renderErrorInModal(error as Error, "You have been rate limited. Please login to increase your limits."); // @DEV: user another method to render the modal not as an error
-      displayPopupMessage({
-        modalHeader: "GitHub API rate limit exceeded.",
-        modalBody: "You have been rate limited. Please login to increase your limits.",
-        isError: false,
-      });
-      gitHubLoginButton?.classList.add("highlight");
+      throw error;
```
gentlementlegen, line 70: @0x4007 Had to change this to bubble up the error, because previous code logic would display the error rate limit whatever the error was, which is wrong on a 500 network error for example.

cypress.config.ts
```diff
@@ -11,21 +11,6 @@ export default defineConfig({
   },
   viewportHeight: 900,
   viewportWidth: 1440,
-  env: readEnvironmentVariables(),
```
0x4007, line 14: This is useful you should consider leaving it in

cypress/e2e/devpool.cy.ts
```diff
@@ -175,36 +185,45 @@ describe("DevPool", () => {
...
+          user: githubUser,
+        })
+      );
+    }).as("githubPage");
     cy.visit("/");
     // Check that there is no text field visible for sorting
     cy.get("#filter").should("not.be.visible");
     cy.get("#github-login-button").click();
-    cy.origin("https://github.com/login", () => {
-      const username = Cypress.env("UBIQUIBOT_GITHUB_USERNAME");
```
0x4007, line 189: Are these environment variables not needed anymore?

src/home/fetch-github/fetch-issues-preview.ts
0x4007: Well then you should catch and display the specified error for rate limiting (I think it's a 403?) 

Otherwise the raw errors don't look as presentable.

static/style/inverted-style.css
```diff
@@ -688,4 +688,9 @@
     align-items: center;
     margin-right: 8px;
   }
+  .github-avatar-default {
+    width: 48px;
+    height: 48px;
+    background-color: #7f7f7f;
```
0x4007:
```suggestion
    background-color: #808080;
```

cypress/e2e/devpool.cy.ts
gentlementlegen: No, the login is now faked to avoid tests crashing due to missing environment variables.

cypress.config.ts
gentlementlegen: I can leave it but it will be a function returning empty content, is that fine?

src/home/fetch-github/fetch-issues-preview.ts
gentlementlegen: The full logic within the `catch` clause is:
```TypeScript
    if (!!error && typeof error === "object" && "status" in error && error.status === 403) {
      await handleRateLimit(octokit, error as RequestError);
    } else {
      throw error;
    }
```

So I think we are fine. I tried and get a nice error on `403` and a generic one for the rest.

cypress.config.ts
0x4007: I suppose we should remove useless logic!

gentlementlegen: Well these are not needed for now since it's mocked. We can always find the logic back within the Git history if needed, so will remove it for now.

0x4007 (approved):

PR #62, opened 2024-06-11
fix: the displayed name for the user falls back to its login if empty

### LINKED ISSUE #61: The user name is not displayed for user who didn't set a name
`name` is not required for a GitHub user, it can be left blank, which results in the UI missing the name of the user.

image

Maybe it would be better to use the `login` field which is required and cannot be left blank for GitHub users.

See also: https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user

gentlementlegen:
Resolves #61

### DIFF
--- a/cypress/e2e/devpool.cy.ts
+++ b/cypress/e2e/devpool.cy.ts
@@ -259,7 +259,7 @@ describe("DevPool", () => {
       name: undefined,
     };
     window.localStorage.setItem(
-      "sb-wfzpewmlyiozupulbuur-auth-token",
+      `sb-${Cypress.env("SUPABASE_STORAGE_KEY")}-auth-token`,
       JSON.stringify({
         provider_token: "token",
         access_token: "token",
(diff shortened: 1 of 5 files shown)


gitcoindev (approved):
Change looks straightforward, I reviewed the code. Thank you for adding Cypress tests that cover this scenario, tests also passing in CI.

cypress/e2e/devpool.cy.ts
```diff
@@ -252,4 +252,34 @@ describe("DevPool", () => {
...
     });
   });
+
+  it("Displayed user name should fall back to login when its name is empty", () => {
+    const userWithoutName = {
+      ...githubUser,
+      name: undefined,
+    };
+    window.localStorage.setItem(
+      "sb-wfzpewmlyiozupulbuur-auth-token",
```
0x4007: Seems wrong to hardcode this. Shouldn't it be an environment variable?

0x4007 (approved):

PR #68, opened 2024-06-24
fix: filters are visible on small screens

### LINKED ISSUE #67: Price filter deleted
Signed in and when the lock button renders it deletes the price filter button. @gentlementlegen i think this might be a mistake from when you added this feature. Can't investigate now because on phone. 

image

gentlementlegen:
Resolves #67

### DIFF
--- a/cypress/e2e/devpool.cy.ts
+++ b/cypress/e2e/devpool.cy.ts
@@ -282,4 +282,33 @@ describe("DevPool", () => {
     cy.visit("/");
     cy.get("#authenticated > .full").should("have.text", "octocat");
   });
+
+  it.only("Should display filters on small devices", () => {
+    cy.viewport("iphone-x");
+    cy.intercept("https://api.github.com/user", {
+      statusCode: 200,
+      body: githubUser,
+    }).as("getUser");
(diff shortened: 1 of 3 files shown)


gentlementlegen:
The filters now scroll on `x` axis if the screen is too small:
image

0x4007:
> The filters now scroll on `x` axis if the screen is too small: image

This already is implemented, except it should be the entire bar?

0x4007 (approved):

cypress/e2e/devpool.cy.ts
```diff
@@ -282,4 +282,33 @@ describe("DevPool", () => {
...
+    cy.intercept("https://api.github.com/", (req) => {
+      req.headers["x-oauth-scopes"] = "repo";
+      req.reply({
+        statusCode: 200,
+      });
+    }).as("head");
+    cy.visit("/");
+    cy.get("#authenticated").should("be.visible");
+    cy.get("#augment-access-button").should("be.visible");
+    cy.get('[for="price"]').should("be.visible");
```
0x4007, line 312: That's cool you added this. Thanks.

PR #70, opened 2024-06-24
fix: add state to sorting manager for bottom and top

### LINKED ISSUE #69: Move Filters to Bottom Bar on Portrait
Looks like the filters are barely useable now that we have the side scrolling, and the additional UI on the top right. Intended for mobile, it makes sense to have the frequently used controls near the bottom of the screen so that the app is easier to use. 

IMG_103370966922-1

0x4007:
Resolves https://github.com/ubiquity/work.ubq.fi/issues/69


0x4007:
I always struggle with Cypress

0x4007:
Only doesn't work on my local, the guess is token expiration after one hour and being logged out, but something in our tests is not handling this. 

[iTerm2 Session Jun 25, 2024 at 00:03:39.rtf.zip](https://github.com/user-attachments/files/15957681/iTerm2.Session.Jun.25.2024.at.00.03.39.rtf.zip)

gentlementlegen:
After token expiration, I could not reproduce the problem and still had all the tests up and running, as well as Action tests. This should be investigated, if someone else encounters the issue it could give us more clues.

PR #72, opened 2024-07-12
Fixing broken cache in text filter

### LINKED ISSUE #66: Text Filter Seems Broken
I put a very specific search query, so I expect it to only filter down to a single task. However it seems to be broken. 

It could be related to an invalid issue preview in my cache, as you can see that the error was thrown. 

work ubq fi_ (1)

Sadaf-A:
Resolves #66

### DIFF
--- a/src/home/fetch-github/fetch-and-display-previews.ts
+++ b/src/home/fetch-github/fetch-and-display-previews.ts
@@ -16,39 +16,43 @@ export type Options = {
 };
 
 export async function fetchAndDisplayPreviewsFromCache(sorting?: Sorting, options = { ordering: "normal" }) {
-  let _cachedTasks = getLocalStore(GITHUB_TASKS_STORAGE_KEY) as TaskStorageItems;
-  const _accessToken = await getGitHubAccessToken();
+  try {
+    let _cachedTasks = getLocalStore(GITHUB_TASKS_STORAGE_KEY) as TaskStorageItems;
+    const _accessToken = await getGitHubAccessToken();
 
-  // Refresh the storage if there is no logged-in object in cachedTasks but there is one now.
-  if (_cachedTasks && !_cachedTasks.loggedIn && _accessToken) {
-    localStorage.removeItem(GITHUB_TASKS_STORAGE_KEY);
+    // Refresh the storage if there is no logged-in object in cachedTasks but there is one now.
+    if (_cachedTasks && !_cachedTasks.loggedIn && _accessToken) {
+      localStorage.removeItem(GITHUB_TASKS_STORAGE_KEY);
+      return fetchAndDisplayPreviewsFromNetwork(sorting, options);
+    }
+
+    // If previously logged in but not anymore, clear cache and fetch from network.
+    if (_cachedTasks && _cachedTasks.loggedIn && !_accessToken) {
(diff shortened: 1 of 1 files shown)


0x4007 (changes requested):
Fix your formatting

0x4007 (commented):
Actually this might be correct if you just moved the whole thing in the block. Although that doesn't seem optimal. Can you move only the relevant part inside the try catch please?

Telling CI to do a deploy so I can test now.

0x4007:
As it turns out I can't login and test from my phone. Hope somebody else can take over!

Sadaf-A:
@0x4007 how does it look now?

0x4007 (approved):

0x4007:
I didn't test that it works but the code looks good.

Sadaf-A:
@Keyrxng could you please provide your inputs here?

Keyrxng:
> I put a very specific search query, so I expect it to only filter down to a single task. However it seems to be broken.

> It could be related to an invalid issue preview in my cache, as you can see that the error was thrown.

@Sadaf-A as far as I understand the spec the following needs to be resolved:

- When you search for an issue the preview box should be filled with the issue details not `"No task found for preview id 2292677641"` which is happening on the preview deployment as well as locally for me

> so I expect it to only filter down to a single task.

It's unclear to me whether the spec aims to have the listed issues on the left become a single item if you enter a very specific term but the preview box should display similar details as when you manually click on the issue

P.S With PRs that change/fix UI problems remember to include a screenshot of that issue after it has been resolved

Sadaf-A:
> > I put a very specific search query, so I expect it to only filter down to a single task. However it seems to be broken.
> 
> > It could be related to an invalid issue preview in my cache, as you can see that the error was thrown.
> 
> @Sadaf-A as far as I understand the spec the following needs to be resolved:
> 
> * When you search for an issue the preview box should be filled with the issue details not `"No task found for preview id 2292677641"` which is happening on the preview deployment as well as locally for me
> 
> > so I expect it to only filter down to a single task.
> 
> It's unclear to me whether the spec aims to have the listed issues on the left become a single item if you enter a very specific term but the preview box should display similar details as when you manually click on the issue
> 
> P.S With PRs that change/fix UI problems remember to include a screenshot of that issue after it has been resolved

I wrapped the whole function responsible for rendering from cache with a try and catch block so if it threw an error it will render from network. If the error is still being thrown then it could be possible that the error is thrown when we fetch from the network

Keyrxng:
@Sadaf-A I had another look at things.

- The list of issues should reduce down to one item, when the error modal is visible it does not remove items from the list.

- The error box appears when you arrive to the page with a fresh cache and you try to search. If you refresh the page and try again the error will not appear.

- I manually looked through the cached issue IDs (may have been after refresh, you should also check before too) and could not see the ID which is in the error box maybe you'll want to manually confirm this too

- I'd focus on where the error is thrown from and work back from there debugging until you can resolve or elegantly handle it.

> but the preview box should display similar details as when you manually click on the issue

Forgive me here, I mistook the error notification for the issue preview modal because I was using a different screen size. When you search `toy "demo"` you should see only one issue which you can select via keyboard nav or clicking and it'll render the preview modal.

I hope this is of more help to you, sorry if not.

PR #75, opened 2024-07-23
feat: replace google analytics with custom referral tracking and storage

### LINKED ISSUE #74: Reimplement Tracking Code
Due to the ad blockers, the google analytics reliant version of the devrel tracker is not reliable. We will need to fully redo this feature. Likely will need to be fully custom so this requires storage on the backend. 

- Remove existing google analytics code.
- Set up a storage solution 
- Record referrals. 
- Change query parameter to `ref=`

Ekep-Obasi:
Resolves #74

### DIFF
--- a/.github/workflows/supabase-types.yml
+++ b/.github/workflows/supabase-types.yml
@@ -9,7 +9,7 @@ jobs:
     permissions:
       contents: write
     env:
-      PROJECT_REF: "lelwqeowubwpkhjlzsiu"
+      PROJECT_ID: "lelwqeowubwpkhjlzsiu"
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/supabase-types.yml
+++ b/.github/workflows/supabase-types.yml
@@ -33,11 +33,11 @@ jobs:
       - name: Commit files
         if: ${{contains(steps.git_status.outputs.status, ' ')}}
         run: |
-            git config --global user.name 'github-actions'
-            git config --global user.email 'github-actions@github.com'
-            git add supabase/types.ts
-            git commit -m 'chore: update Supabase types'
-            git push
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@github.com'
+          git add supabase/types.ts
+          git commit -m 'chore: update Supabase types'
+          git push
 
       - name: Push changes
         if: ${{contains(steps.git_status.outputs.status, ' ')}}
--- a/.github/workflows/supabase-types.yml
+++ b/.github/workflows/supabase-types.yml
@@ -2,6 +2,8 @@ name: Update database types
 
 on:
   push:
+    branches:
+      - main
 
 jobs:
   update:
@@ -9,13 +11,14 @@ jobs:
     permissions:
       contents: write
     env:
-      PROJECT_ID: "lelwqeowubwpkhjlzsiu"
+      PROJECT_ID: ${{ secrets.PROJECT_ID }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           persist-credentials: false
           fetch-depth: 0
-      - uses: actions/setup-node@v2.1.5
+
+      - uses: actions/setup-node@v4
         with:
           node-version: 20.10.0
 
@@ -23,24 +26,23 @@ jobs:
         run: yarn install
 
       - name: Generate Supabase types
-        run: npm run update-types
+        run: yarn run update-types
 
-      - name: check for file changes
+      - name: Check for file changes
         id: git_status
         run: |
           echo "status=$(git status -s)" >> $GITHUB_OUTPUT
 
       - name: Commit files
-        if: ${{contains(steps.git_status.outputs.status, ' ')}}
+        if: ${{ contains(steps.git_status.outputs.status, ' ') }}
         run: |
           git config --global user.name 'github-actions'
           git config --global user.email 'github-actions@github.com'
           git add supabase/types.ts
           git commit -m 'chore: update Supabase types'
-          git push
 
       - name: Push changes
-        if: ${{contains(steps.git_status.outputs.status, ' ')}}
+        if: ${{ contains(steps.git_status.outputs.status, ' ') }}
         uses: ad-m/github-push-action@master
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/supabase-types.yml
+++ b/.github/workflows/supabase-types.yml
@@ -1,0 +1,47 @@
+name: Update database types
+
+on:
+  push:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      PROJECT_REF: "lelwqeowubwpkhjlzsiu"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - uses: actions/setup-node@v2.1.5
+        with:
+          node-version: 20.10.0
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Generate Supabase types
+        run: npm run update-types
+
+      - name: check for file changes
+        id: git_status
+        run: |
+          echo "status=$(git status -s)" >> $GITHUB_OUTPUT
+
+      - name: Commit files
+        if: ${{contains(steps.git_status.outputs.status, ' ')}}
+        run: |
+            git config --global user.name 'github-actions'
+            git config --global user.email 'github-actions@github.com'
+            git add supabase/types.ts
+            git commit -m 'chore: update Supabase types'
+            git push
+
+      - name: Push changes
+        if: ${{contains(steps.git_status.outputs.status, ' ')}}
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.ref }}
--- a/cypress/e2e/devpool.cy.ts
+++ b/cypress/e2e/devpool.cy.ts
@@ -291,6 +291,49 @@ describe("DevPool", () => {
     cy.get("#filter-top").should("be.visible");
   });
 
+  it("Should save referral code to Supabase referral table upon login", () => {
+    const referralCode = "test-referral-code";
+
+    cy.intercept("https://api.github.com/user**", (req) => {
+      req.reply({
+        statusCode: 200,
+        body: githubUser,
+      });
+    }).as("getUser");
+
+    cy.intercept("POST", "https://lelwqeowubwpkhjlzsiu.supabase.co/rest/v1/referrals", (req) => {
+      expect(req.body).to.have.property("referral_code", referralCode);
+      expect(req.body).to.have.property("dev_github", `${githubUser.login}|${githubUser.id}`);
+      req.reply({
+        statusCode: 201
+      });
+    }).as("saveReferral");
+
+    const loginUrl = `?ref=${referralCode}`;
+    cy.visit(loginUrl);
+    cy.get("#github-login-button").click();
+
+    cy.intercept("https://github.com/login**", (req) => {
+      req.reply({
+        statusCode: 200,
+      });
+
+      window.localStorage.setItem(
+        `sb-${Cypress.env("SUPABASE_STORAGE_KEY")}-auth-token`,
+        JSON.stringify({
+          provider_token: "token",
+          access_token: "token",
+          token_type: "bearer",
+          user: githubUser,
+        })
+      );
+    }).as("githubLogin");
+
+    cy.visit("/");
+
+    cy.wait("@saveReferral");
+  });
+
   describe("Display error modal", () => {
     it("should display an error modal when fetching issue previews fails on page load", () => {
       cy.intercept("GET", "https://api.github.com/repos/ubiquity/devpool-directory/issues*", {
--- a/cypress/e2e/devpool.cy.ts
+++ b/cypress/e2e/devpool.cy.ts
@@ -205,6 +205,49 @@ describe("DevPool", () => {
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
   });
 
+  it("Should save referral code to Supabase referral table upon login", () => {
+    const referralCode = "test-referral-code";
+
+    cy.intercept("https://api.github.com/user**", (req) => {
+      req.reply({
+        statusCode: 200,
+        body: githubUser,
+      });
+    }).as("getUser");
+
+    cy.intercept("POST", `${Cypress.env("SUPABASE_URL")}/rest/v1/referrals`, (req) => {
+      expect(req.body).to.have.property("referral_code", referralCode);
+      expect(req.body).to.have.property("dev_github", `${githubUser.login}|${githubUser.id}`);
+      req.reply({
+        statusCode: 201,
+      });
+    }).as("saveReferral");
+
+    const loginUrl = `?ref=${referralCode}`;
+    cy.visit(loginUrl);
+    cy.get("#github-login-button").click();
+
+    cy.intercept("https://github.com/login**", (req) => {
+      req.reply({
+        statusCode: 200,
+      });
+
+      window.localStorage.setItem(
+        `sb-${Cypress.env("SUPABASE_STORAGE_KEY")}-auth-token`,
+        JSON.stringify({
+          provider_token: "token",
+          access_token: "token",
+          token_type: "bearer",
+          user: githubUser,
+        })
+      );
+    }).as("githubLogin");
+
+    cy.visit("/");
+
+    cy.wait("@saveReferral");
+  });
+
   it("User can log in", () => {
     cy.intercept("https://api.github.com/user**", (req) => {
       req.reply({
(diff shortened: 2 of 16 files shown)


0x4007 (commented):
Code looks good. @gentlementlegen you're pretty good with supabase what do you think about this?

gentlementlegen:
Would be nice to:
- have CI tests pass
- have an Action that generates Supabase types and commits them so contributors that do not have access to the database can have up to date types
- add related tests
- use Ubiquity's logger instead of console logs

0x4007:
> Would be nice to:
> 
> * have CI tests pass

None of the code changes stand out to me as causing Cypress to fail. Perhaps the test is unreliable?

gentlementlegen:
Error is here
```
  1) DevPool
       Items can be sorted - bottom row - portrait/mobile:
     CypressError: Timed out retrying after 4000ms: `cy.scrollTo()` failed because this element is not scrollable:

`<div id="filters-bottom" class="filters-container">...</div>`

Make sure you're targeting the correct element or use `{ensureScrollable: false}` to disable the scrollable check.
```

Maybe this error was here before.

cypress/e2e/devpool.cy.ts
```diff
@@ -250,6 +250,49 @@ describe("DevPool", () => {
...
+    const referralCode = "test-referral-code";
+
+    cy.intercept("https://api.github.com/user**", (req) => {
+      req.reply({
+        statusCode: 200,
+        body: githubUser,
+      });
+    }).as("getUser");
+
+    cy.intercept("POST", "https://lelwqeowubwpkhjlzsiu.supabase.co/rest/v1/referrals", (req) => {
```
gentlementlegen: Ideally this should not be hardcoded.

cypress/e2e/devpool.cy.ts
```diff
@@ -199,7 +199,7 @@ describe("DevPool", () => {
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
     cy.get('[for="priority-bottom"]').click();
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
-    cy.get("#filters-bottom").scrollTo("right");
+    cy.get("#filters-bottom").scrollTo("right", {ensureScrollable: false});
```
gentlementlegen: Doesn't seem to respect prettier rules.

Ekep-Obasi: Thanks @gentlementlegen, let me do the changes, Could you approve the workflows

gentlementlegen: I think the workflows are working for each commit?

0x4007: > I think the workflows are working for each commit?

I ran it manually

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
...
+on:
+  push:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      PROJECT_REF: "lelwqeowubwpkhjlzsiu"
```
gentlementlegen: Should come from the environment

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
+name: Update database types
+
+on:
+  push:
```
gentlementlegen, line 4: I think this should happen only on the main branch, so it's the latest types and it doesn't bombard the repo with commits.

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
...
+    permissions:
+      contents: write
+    env:
+      PROJECT_REF: "lelwqeowubwpkhjlzsiu"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - uses: actions/setup-node@v2.1.5
```
gentlementlegen: Why this version specifically?

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      PROJECT_REF: "lelwqeowubwpkhjlzsiu"
+    steps:
+      - uses: actions/checkout@v2
```
gentlementlegen:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Commit files
+        if: ${{contains(steps.git_status.outputs.status, ' ')}}
+        run: |
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@github.com'
+          git add supabase/types.ts
+          git commit -m 'chore: update Supabase types'
+          git push
+
+      - name: Push changes
```
gentlementlegen, line 44: You use this step but the latest command of the previous step seems to be `git push`?

supabase/tables/base.ts
```diff
@@ -0,0 +1,12 @@
+import { SupabaseClient } from "@supabase/supabase-js";
+import { Logs } from "ubiquibot-logger";
+
+export class SuperBase {
+  protected supabase: SupabaseClient;
+  public logger: Logs;
+
+  constructor(supabase: SupabaseClient) {
+    this.supabase = supabase;
+    this.logger = {} as Logs;
```
gentlementlegen: Why not `new Logger()`?

.github/workflows/supabase-types.yml
```diff
@@ -0,0 +1,47 @@
...
+          fetch-depth: 0
+      - uses: actions/setup-node@v2.1.5
+        with:
+          node-version: 20.10.0
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Generate Supabase types
+        run: npm run update-types
```
gentlementlegen: Probably should use `yarn`

supabase/tables/base.ts
```diff
@@ -0,0 +1,12 @@
+import { SupabaseClient } from "@supabase/supabase-js";
+import { Logs } from "ubiquibot-logger";
+
+export class SuperBase {
```
0x4007:
```suggestion
export class SupabaseClient {
```

Maybe this is more accurate.

supabase/config.toml
```diff
@@ -0,0 +1,178 @@
...
+client_id = ""
+# DO NOT commit your OAuth provider secret to git. Use environment variable substitution instead:
+secret = "env(SUPABASE_AUTH_EXTERNAL_APPLE_SECRET)"
+# Overrides the default auth redirectUrl.
+redirect_uri = ""
+# Overrides the default auth provider URL. Used to support self-hosted gitlab, single-tenant Azure,
+# or any other third-party OIDC providers.
+url = ""
+# If enabled, the nonce check will be skipped. Required for local sign in with Google auth.
+skip_nonce_check = false
```
0x4007:
```suggestion
```
Perhaps there are some unnecessary configs.

supabase/tables/base.ts
Ekep-Obasi: I avoided `new Logger()` to keep the constructor simple and avoid unnecessary configuration which are not optional. Instead, I initialized logger as an empty object for simplicity.

supabase/config.toml
Ekep-Obasi: I’m not very familiar with this, but I could clean up some of the comments

package.json
```diff
@@ -17,7 +17,8 @@
     "prepare": "husky install",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:headed": "cypress run --headed"
+    "cypress:headed": "cypress run --headed",
+    "update-types": "npx supabase gen types --lang=typescript --project-id $(grep -oP '(?<=^PROJECT_ID=).+' .env) > supabase/types.ts"
```
gentlementlegen: `grep` is not cross environment friendly, and does this handle local / remote environments?

package.json
```diff
@@ -52,6 +53,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^3.1.0",
     "tsx": "^4.6.0",
+    "ubiquibot-logger": "^0.3.5",
```
gentlementlegen: https://www.npmjs.com/package/@ubiquity-dao/ubiquibot-logger should be the proper package.

package.json
Ekep-Obasi: Yeah that true, It'll be preferable to write a script

gentlementlegen: Using a script is fine but maybe overkill, you can just use `cross-env` and `dotenv` to load the variable within the command line (and it works on any environment). Also, would be nice to have a `local` and `remote` command to generate the types either locally with a local db or remotely targeting a deployed database.

Ekep-Obasi:
@gentlementlegen, How are the changes looking now? Are there other areas where I could make improvements?

supabase/tables/base.ts
gentlementlegen: So are the inheritors responsible to initialize the logger? What happens if a class does not initialize it?

package.json
```diff
@@ -17,7 +17,8 @@
     "prepare": "husky install",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:headed": "cypress run --headed"
+    "cypress:headed": "cypress run --headed",
+    "update-types": "dotenv -e .env -- sh -c 'npx supabase gen types --lang=typescript --project-id $PROJECT_ID > supabase/types.ts'"
```
gentlementlegen: `sh` is also not cross platform. Please have a look at https://www.npmjs.com/package/cross-env

supabase/config.toml
gentlementlegen, line 1: Does this imply we need to spin up a new Supabase database?

package.json
Ekep-Obasi: From what I see here,`cross-env` is good for setting environment variables in a cross-platform way but does not handle `.env` files.

gentlementlegen: Correct which is why you should combine `cross-env-shell` to run the script and `dotenv` to load the environment file.

supabase/tables/base.ts
Ekep-Obasi: You're right. Inheritors need to initialize the logger, or it will stay as an empty object and could cause issues.

supabase/config.toml
Ekep-Obasi: yeah, I think so

package.json
Ekep-Obasi: I went through `cross-env-shell` package, I could not really find anything helpful https://www.npmjs.com/package/cross-env-shell.

But I went for another similar package `dotenv-cli`, Using it with `bash -c` in GitHub Actions, which runs on Ubuntu should work

Ekep-Obasi:
@gentlementlegen Could you please approve the workflows?

Ekep-Obasi:
Hey @gentlementlegen, @0x4007, @Keyrxng,

I wanted to follow up on my pull request from last week. When you get a chance, I would appreciate it if someone could take a look and approve it if everything looks good.

supabase/seed.sql
Keyrxng, line 1: shouldn't this contain a script to setup the DB? Or there should be a migration script instead but I don't see the value in an empty file here

supabase/types.ts
Keyrxng, line 1: We usually avoid committing these I'm sure. Is this strictly needed? Could a simple type be used instead to represent the table?

supabase/config.toml
Keyrxng, line 1: Can this config be stripped down to only what we need to use?

src/home/devrel-tracker.ts
```diff
@@ -1,22 +1,30 @@
...
 export function initiateDevRelTracking() {
-  const oldDevRelCode = localStorage.getItem("devRel");
+  const oldDevRelCode = localStorage.getItem("ref");
   if (!oldDevRelCode) {
     const urlParams = new URLSearchParams(window.location.search);
-    const devRelCode = urlParams.get("devRel");
+    const devRelCode = urlParams.get("ref");
     if (devRelCode) {
-      localStorage.setItem("devRel", devRelCode);
+      localStorage.setItem("ref", devRelCode);
```
Keyrxng, line 10: > One thing that I Just noticed is that in the URL https://work.ubq.fi/?devRel=XYZ the query parameter devRel is case-sensitive. The R in devRel must be capitalized. It would have been nice to have it case-insensitive but this is what is working now. 

A comment from the original implementation PR @gentlementlegen is this something we should consider? I'm unsure what the referral url schema looks like beyond `ref=`

Keyrxng (changes requested):
The localStorage and posting to Supabase logic works

I setup a new database and had to disable RLS to allow me to push. RLS should probably be defined for this table right?

I'd suggest for the authenticated user and that they can only edit their own entry, probably reads should be disabled as well

This can be done as part of the migration/setup script

supabase/types.ts
Keyrxng: @gentlementlegen rfc

edit:


I just saw [your comment](https://github.com/ubiquity/work.ubq.fi/pull/75#issuecomment-2246858755) and to follow on

> Does this imply we need to spin up a new Supabase database?
> yeah, I think so

so are we spinning up a new DB just for referrals or just adding a table to the main DB? Either way I'd have thought if we are rebuilding the referral system from scratch then there would be no breaking changes to come for a long time and simple types would be better

Just my opinion, I think the Supabase types are unattractive and can be overwhelming but them being dynamic is ofc a bonus

Keyrxng:
@0x4007 with what I assume is more freedom with what can be stored in the DB vs GTAG events should this be given some thought as to how we store things

> Set up a storage solution
> Record referrals.

e.g do we need `dev_github = "username|userId"` or does it improve things if they are separate rows? Could the primaryKey be the `userId` to easily enforce uniqueness/once-per and authenticated-user RLS edits?

Perhaps it's best that records cannot be edited or seen at all once posted to the DB?

I'll admit I'm not fantastic with architecting backend storage solutions at the best of times but definitely not when practically off the top of my head

0x4007:
All data should be in different columns as much as possible to make querying it as accurate as possible.

src/home/devrel-tracker.ts
0x4007: I think `ref` is sufficient for our needs here.

Keyrxng: sorry @0x4007 I meant what comes after `ref=`, the code or string or whatever the schema is, is case sensitivity a concern there? I presume no with the above but I haven't seen any campaign link generating first-hand so I wanted to check, ty.

supabase/types.ts
gentlementlegen: I think it is good and bad to commit these, but so far we always did.
Pros:
- any developer can clone and use
- maybe developers do not have access to the DB and it's nice to have the types without spinning up a local db

Cons:
- pollutes the repo
- easily can merge conflict
- knip will complain on missing types

I think the best scenario is to have an Action generating the types on any push to the main branch as it ensures that the deployed db and the code are up to date with each other. Either way there has to be some script generating these types, either during build or during a push to the main branch.

package.json
```diff
@@ -17,7 +17,7 @@
     "prepare": "husky install",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:headed": "cypress run --headed"
```
gentlementlegen, line 20: Why was cypress removed?

package.json
```diff
@@ -17,7 +17,7 @@
...
+    "update-types": "dotenv -e .env -- bash -c 'npx supabase gen types --lang=typescript --project-id $PROJECT_ID --schema public > supabase/types.ts'"
```
gentlementlegen: Thank you for this. But please make it cross platform (this would not work outside of bash environments). Please refer to `cross-env` or equivalent.

package.json
```diff
@@ -41,7 +43,7 @@
     "@types/node": "^20.10.0",
     "@typescript-eslint/eslint-plugin": "^6.13.1",
     "@typescript-eslint/parser": "^6.13.1",
-    "cypress": "13.11.0",
+    "cypress": "^13.11.0",
```
gentlementlegen:
```suggestion
    "cypress": "13.11.0",
```

supabase/tables/base.ts
```diff
@@ -0,0 +1,12 @@
+import { SupabaseClient } from "@supabase/supabase-js";
+import { Logs, LOG_LEVEL, LogLevel } from "@ubiquity-dao/ubiquibot-logger";
+
+export class SupaBase {
```
gentlementlegen:
```suggestion
export class Supabase {
```
Except if this names collides with something.

package.json
Ekep-Obasi: my bad, I must have removed this by mistake

package.json
```diff
@@ -17,7 +17,8 @@
     "prepare": "husky install",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:headed": "cypress run --headed"
+    "cypress:headed": "cypress run --headed",
+    "update-types": "cross-env dotenv -e .env -- bash -c 'npx supabase gen types --lang=typescript --project-id $PROJECT_ID --schema public > supabase/types.ts'"
```
Keyrxng, line 21: Without a migration file or a way to setup a DB programmatically, if a new contributor wanted to use their own DB they'd need to root around your code/types and manually create it first.

Failing that they'd erase your types and the logic would fail as the DB would not be ready to accept the data. They'd also have to create RLS policies too, which would typically be a part of the migration file.

What is your solution to this problem?

package.json
```diff
@@ -30,7 +31,10 @@
     "@octokit/request-error": "^6.1.0",
     "@octokit/rest": "^20.0.2",
     "@supabase/supabase-js": "^2.39.0",
+    "@ubiquity-dao/ubiquibot-logger": "^1.3.0",
```
Keyrxng, line 34: can you bump this to `1.3.1` please?

supabase/types.ts
Keyrxng, line 1: @gentlementlegen double checking here that we are definitely committing this file?

If we use the main DB it exposes everything and we usually avoid that. If we spin up a DB dedicated to referrals it means more infra to manage and less interoperable with the main DB.

supabase/config.toml
Keyrxng: This is still unresolved. 

Ask yourself what features of this have you been using while developing? Out of curiosity, are you using a local or remote Supabase instance and did you setup a Github app for logging in and testing with or did you use the ubiquity DB for auth? 

How is the app connected to Supabase and what auth strategy does the UI implement? What is it we are using the storage api and buckets for, are they required? Do we login with twillio or sms or email? Do we need S3 env keys? 

Likely most, if not all of these can be removed. I'm no expert with Supabase but what do we do with this file within the context of this repo/feature? Obviously it's for configuring a DB but where/how is it used?

Keyrxng (changes requested):
Thinking in terms of future contributions, they will need to use their own DB instance as the production DB should have RLS that prevents dangerous actions on it which would make testing anything to do with the feature difficult if using the main DB details.

This means manually creating a new DB/new tables every time unless we have a migration file included as part of this PR.

I think the readme should be updated to include steps on how a new contributor would setup the DB and what's needed as part of this new feature.

supabase/types.ts
gentlementlegen: We are committing this to other projects, and we need it for Knip not to complain. You can always introspect Supabase afaik, as long as you provide an ANON key (try it out here https://supabase-schema.vercel.app/). If so then it should be fine.

PR #77, opened 2024-08-15
Only display tasks with price labels (or without, depending on env)

### LINKED ISSUE #115: Add support for proposals on `work.ubq.fi`
Related https://github.com/ubiquity/work.ubq.fi/issues/54

There's a lot of unaccounted for/half completed proposals and it will be best to have a dedicated view to sort and browse. 

I recreated the task here because this involves work across several repositories. 

- The first thing that needs to be addressed is to upgrade the [backend](https://github.com/ubiquity/devpool-directory/issues). Perhaps we can make a new repository with a better name vs `devpool-directory` although it isn't clear to me what the name should be yet. Perhaps we can call it "organization issue aggregator" as sort of a general purpose name. It should list every issue in our organization, and include the price tag if there is a price. If not, then no price tag should be listed
- Next, on work.ubq.fi it should only filter for issues with price tags. 
- Then we can make a separate view that will filter for issues without price tags. It will allow us to keep the logic consolidated together. 
- Lastly we should figure something about syncing https://github.com/ubiquity/devpool-directory and https://github.com/ubiquity/devpool-directory-private I hope it can be rolled into this task but given that there is no details around how to implement, it might need to be broken off into a separate task.

xno-miner:
Resolves https://github.com/ubiquity/.github/issues/115

Filter issues with price label when loading issue previews in `src/home/fetch-github/fetch-issues-preview.ts`.


xno-miner:
This part is done (it works for me locally at least).

The amount of issues that can be fetched with a single request is 30 (default) or 100 (max). There are currently 92 issues in the devpool (including those without price label) so I implemented pagination to avoid future problems with the website.

package.json
```diff
@@ -64,5 +64,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
0x4007:
```suggestion
```

Remove

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -48,13 +48,27 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
...
+        repo: "devpool-directory",
+        state: "open",
+        per_page: 100,
+        page: page,
+      });
+
+      data = [...data, ...pageData];
+
+      if (pageData.length < 100 ) {
+        break;
```
0x4007: Why? Pretty sure there is a built in method to handle pagination in Octokit. Please redo.

xno-miner: didn't know that. fixed

xno-miner:
I added an `RFC` enviromental variable. For the `false` value it works just as before. It normally wouldn't work for `true` value yet, but it almost does, because of the issues i accidentaly created. The only problem is it doesn't display any labels. This is becase the issues it's displaying don't have any (they're the ones i accidentally added with my bot). For the labels to be displayed, all issues in `devpool-directory` created by testbot6 need to be removed. After that, once the backend will be merged and the issues will be synced, the labels will be displayed correctly.

rndquu:
@xno-miner Could you fix the tests https://github.com/ubiquity/work.ubq.fi/actions/runs/10434585610/job/28922040659?pr=77 ?

xno-miner:
> @xno-miner Could you fix the tests https://github.com/ubiquity/work.ubq.fi/actions/runs/10434585610/job/28922040659?pr=77 ?

Ok

xno-miner:
> @xno-miner Could you fix the tests

Seems like the tests were also failing in the main branch and have been fixed here: https://github.com/ubiquity/work.ubq.fi/pull/77/commits/f14181ae0aa962a569efce60707e3d808cb1bd4c. I have added this commit to the branch.

0x4007:
@rndquu do you mind seeing this rfc project reviews through?

build/esbuild-build.ts
```diff
@@ -29,6 +29,7 @@ export const esBuildContext: esbuild.BuildOptions = {
     SUPABASE_STORAGE_KEY: generateSupabaseStorageKey(),
     commitHash: execSync(`git rev-parse --short HEAD`).toString().trim(),
     NODE_ENV: process.env.NODE_ENV || "development",
+    IS_RFC: JSON.parse(process.env.RFC!)
```
rndquu: It makes sense to provide a default value here in case `process.env.RFC` is empty, like `process.env.RFC || false`. This way we make it backwards compatible with already deployed version of `work.ubq.fi`.

cypress/e2e/devpool.cy.ts
```diff
@@ -199,7 +199,6 @@ describe("DevPool", () => {
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
     cy.get('[for="priority-bottom"]').click();
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
-    cy.get("#filters-bottom").scrollTo("right");
```
rndquu: Why was this line removed?

src/home/fetch-github/fetch-issues-preview.ts
```diff
@@ -71,7 +77,7 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
     }
   }
 
-  const tasks = freshIssues.map((preview: GitHubIssue) => ({
+  const tasks = freshIssues.filter((issue: GitHubIssue) => (issue.labels as GitHubLabel[]).some((label) => label.name.includes("Pricing")) == !IS_RFC).map((preview: GitHubIssue) => ({
```
rndquu, line 80: How `IS_RFC` gets a value here? I mean that we define it [here](https://github.com/ubiquity/work.ubq.fi/blob/334442e26383130dbc6b32a64a34dfcd72b6deb4/src/home/fetch-github/fetch-issues-preview.ts#L1) but don't assign it a value anywhere. Is it working?

rndquu (changes requested):
@xno-miner 

1. Pls resolve code conflicts 
2. Getting this error when setting `RFC=true` in the `.env` file:
Screenshot 2024-08-27 at 18 26 44

Pls fix

cypress/e2e/devpool.cy.ts
0x4007: I think @gentlementlegen removed it because it's an unreliable test. It fails intermittently

xno-miner:
>     2. Getting this error when setting `RFC=true` in the `.env` file:
> 
> 
> Screenshot 2024-08-27 at 18 26 44
> 
> Pls fix

The same error exists in the deployed version on https://work.ubq.fi. The error exists because some issue links include `www.` which causes the regex to not match. I will create a pull request in this repository separate from this one so it can be merged separately quicker.

src/home/fetch-github/fetch-issues-preview.ts
xno-miner: env variables are passed in at build time

source: comments at the top of `src/home/rendering/render-github-login-button.ts`

xno-miner:
> The same error exists in the deployed version on https://work.ubq.fi. The error exists because some issue links include `www.` which causes the regex to not match. I will create a pull request in this repository separate from this one so it can be merged separately quicker.

I have created the PR: https://github.com/ubiquity/work.ubq.fi/pull/86

build/esbuild-build.ts
xno-miner: https://github.com/ubiquity/work.ubq.fi/pull/77/commits/f27d1d7c8ad76eff8ea968018b76e5c875e6fbf5

xno-miner:
Still waiting for review, @rndquu.
Here, as well as in https://github.com/ubiquity/devpool-directory/pull/1250

.env.example
```diff
@@ -1,2 +1,3 @@
 SUPABASE_URL=
 SUPABASE_ANON_KEY=
+RFC=false
```
0x4007:
```suggestion
RFC=false 
```

0x4007:
Is the idea to fork this repo and run rfc.ubq.fi off of it? They will need to be hosted on two separate organizations for a fork to be possible. 

I realize we can't QA test this with our deployer because you're setting RFC to false in the [build script](https://github.com/ubiquity/work.ubq.fi/blob/6687fefafc6c42731f102f908fbb0550f8e25648/.github/workflows/build.yml#L33). 

Deploys: https://github.com/ubiquity/work.ubq.fi/pull/77#issuecomment-2291946789

0x4007:
I realize we can't QA test this with our deployer because you're setting RFC to false in the build script. 

https://github.com/ubiquity/work.ubq.fi/pull/77#issuecomment-2291946789

.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         env: # Set environment variables for the build
           SUPABASE_URL: "https://wfzpewmlyiozupulbuur.supabase.co"
           SUPABASE_ANON_KEY: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6IndmenBld21seWlvenVwdWxidXVyIiwicm9sZSI6ImFub24iLCJpYXQiOjE2OTU2NzQzMzksImV4cCI6MjAxMTI1MDMzOX0.SKIL3Q0NOBaMehH0ekFspwgcu3afp3Dl9EDzPqs1nKs"
-
+          RFC: "false"
```
0x4007:
```suggestion
          RFC: true
```

0x4007:
## QA Test Results

### Context

I merged the "backend" changes (to sync every issue basically) and now work.ubq.fi currently displays everything, even if they don't have prices. 

image

### Requested Changes

- This makes me think: the environment variable approach seems wrong. 
- Why don't you just add a checkbox toggle/filter "Tasks" and when clicked it switches to "Proposals" or something?
- This is intended to hide all the issues without a `Price` label.

### "Backend" Delta Summary

58 open ➡️ 143 open...

```
Run npx tsx ./scripts/delete-unauthorized-issues.ts
  
Deleting issue https://github.com/ubiquity/devpool-directory/issues/1[10](https://github.com/ubiquity/devpool-directory/actions/runs/11041825775/job/30672939774#step:8:10)9 created by bot ubiquibot[bot]
Deleting issue https://github.com/ubiquity/devpool-directory/issues/817 created by bot ubiquibot[bot]
Deleting issue https://github.com/ubiquity/devpool-directory/issues/816 created by bot ubiquibot[bot]
Deleting issue https://github.com/ubiquity/devpool-directory/issues/815 created by bot ubiquibot[bot]
Deleting issue https://github.com/ubiquity/devpool-directory/issues/741 created by bot ubiquibot[bot]
Deleting issue https://github.com/ubiquity/devpool-directory/issues/589 created by bot ubiquibot[bot]
```

137 open net result. 

### Screenshot 

It looks like it works as expected with RFC env var! How did you test/develop this? Perhaps you could have shown us screenshots and proof on your end? Apologies for the slow review period but that certainly would have sped things along!

e93b0680 devpool-directory-ui pages dev_

### Remarks 

### Archived Repositories

- Just realized we have another task open for it so we can let it slide for this pull https://github.com/ubiquity/devpool-directory-bounties/issues/32
- Example: https://github.com/ubiquity/launch-party/issues/4

PR #79, opened 2024-08-18
fix: reimpliment highlight functionality on rate limit

### LINKED ISSUE #78: Rate Limited Login Button No Longer Highlighted
image

Upon rate limiting I made it so the login button draws attention to itself by cycling with sort of a blinking highlight animation. This no longer exists and should be reverted.

jordan-ae:
Resolves #78
Screenshot from 2024-08-18 18-56-39

### DIFF
--- a/cypress/e2e/devpool.cy.ts
+++ b/cypress/e2e/devpool.cy.ts
@@ -199,7 +199,6 @@ describe("DevPool", () => {
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
     cy.get('[for="priority-bottom"]').click();
     cy.get('div[id="issues-container"]').children().should("have.length", 2);
-    cy.get("#filters-bottom").scrollTo("right");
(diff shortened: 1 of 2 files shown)


0x4007:
@gentlementlegen This cypress error seems to reoccur intermittently. Can you look into this? Doesn't seem like a good idea to merge commits that aren't passing CI

0x4007 (approved):
I can't git blame easily from mobile but curious why this was removed in the first place.

jordan-ae:
@0x4007 it looks like it was deleted by you when you were re-implementing error toast

gentlementlegen:
@0x4007 This test has been failing for quite some time but it seems that PRs went in regardless. I think it's one line of code to remove in the test, because the bottom bar does not scroll anymore (got changed when FAQ button was added I believe). I will commit in this PR.

---

Should be good to go now. Also tested the last build, the login button shines on desktop and mobile on rate limitation.

PR #81, opened 2024-08-19
fix: paginate issues

### LINKED ISSUE #80: Fetch all devpool issues
The public and private issue fetch calls do not leverage `octokit.paginate` which is contributing to the current task issue onsite.

Keyrxng:
Resolves https://github.com/ubiquity/work.ubq.fi/issues/80

### DIFF
--- a/src/home/fetch-github/fetch-issues-preview.ts
+++ b/src/home/fetch-github/fetch-issues-preview.ts
@@ -48,7 +48,7 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
     hasPrivateRepoAccess = await checkPrivateRepoAccess();
 
     // Fetch issues from public repository
-    const { data: publicResponse } = await octokit.issues.listForRepo({
+    const publicResponse = await octokit.paginate(octokit.issues.listForRepo, {
       owner: "ubiquity",
       repo: "devpool-directory",
       state: "open",
@@ -81,7 +81,7 @@ export async function fetchIssuePreviews(): Promise<TaskNoFull[]> {
   return tasks;
 
   async function fetchPrivateIssues(publicIssues: GitHubIssue[]) {
-    const { data: privateResponse } = await octokit.issues.listForRepo({
+    const privateResponse = await octokit.paginate(octokit.issues.listForRepo, {
       owner: "ubiquity",
       repo: "devpool-directory-private",
       state: "open",


src/home/rendering/render-github-issues.ts
```diff
@@ -17,8 +17,14 @@ export function renderGitHubIssues(tasks: TaskMaybeFull[]) {
...
   const baseDelay = 1000 / 15; // Base delay in milliseconds
-
   for (const task of tasks) {
+    const isBot = task.preview.user?.type === "Bot"
+    const userId = task.preview.user?.id
+
+    if (isBot && userId !== 113181824 && userId !== 76412717 && userId !== 165700353 && userId !== 133917611) {
+      continue;
+    }
+
```
0x4007: Quick fix but we should remember to implement correctly asap.

src/home/rendering/render-github-issues.ts
```diff
@@ -75,11 +81,10 @@ function setUpIssueElement(
...
   issueElement.innerHTML = `
-      <div class="info"><div class="title"><h3>${
-        task.preview.title
-      }</h3></div><div class="partner"><p class="organization-name">${organizationName}</p><p class="repository-name">${repositoryName}</p></div></div><div class="labels">${labels.join(
-        ""
-      )}${image}</div>`;
+      <div class="info"><div class="title"><h3>${task.preview.title
+    }</h3></div><div class="partner"><p class="organization-name">${organizationName}</p><p class="repository-name">${repositoryName}</p></div></div><div class="labels">${labels.join(
+      ""
+    )}${image}</div>`;
```
0x4007: What is this? Can you run the formatter?

Keyrxng: format-on-save I believe, no actual changes were made here, I'll run format.

src/home/rendering/render-github-issues.ts
Keyrxng: I took the 3 IDs at the end from your [comment](https://github.com/ubiquity/devpool-directory-bounties/issues/30#issuecomment-2292647684)

I see I see my mistake

Keyrxng: This error while implementing things as suggested

> You must authenticate with an access token authorized to a GitHub App in order to list installations - https://docs.github.com/rest/apps/installations#list-app-installations-accessible-to-the-user-access-token

So we'd need to store our bot private key here so we can verify that way you proposed for the cron job? 

I think it's overkill for something that will be removed in no time at all once the cron is completed.

rndquu: Is this logic correct? I mean that if an issue in https://github.com/ubiquity/devpool-directory is created by a user then `isBot = false` and the `if (isBot && userId !== 113181824 && userId !== 76412717 && userId !== 165700353 && userId !== 133917611)` expression evaluates to `false` thus rendering an issue by "unauthorized" user. While the expected behavior is to skip rendering such github issues.

Overall it should be simply filtered on the backend instead in order to omit such dirty hacks.

rndquu: I've removed all issues created by `testbot6` in https://github.com/ubiquity/devpool-directory/issues so this check seems to be redundant:
```
    if (isBot && userId !== 113181824 && userId !== 76412717 && userId !== 165700353 && userId !== 133917611) {
      continue;
    }
```

Again, it's better to simply fix it on the backend https://github.com/ubiquity/devpool-directory-bounties/issues/31

Keyrxng: > Overall it should be simply filtered on the backend instead in order to omit such dirty hacks.

Agreed.

> Is this logic correct? 

In terms of preventing bots yeah but I hadn't considered that user's can create issues too

> Again, it's better to simply fix it on the backend https://github.com/ubiquity/devpool-directory-bounties/issues/31

So I should remove these checks here and we address the backend immediately instead

rndquu:
Since there are no more "spam" github issues created by `testbot6` at https://github.com/ubiquity/devpool-directory/issues then it makes sense to 
1. Keep this PR responsible only for the pagination feature 
2. Remove [bot checks](https://github.com/ubiquity/work.ubq.fi/pull/81/files#diff-dd804fd19e44d553ea26928f9e72a6198775a1c3af9605afebd7935c5b57d05aR21-R27) and let the backend [handle](https://github.com/ubiquity/devpool-directory-bounties/issues/31) it

Keyrxng:
cleaned up the commit history a bit and this PR now only adds pagination

rndquu (approved):

PR #83, opened 2024-08-21
Add sync.yml workflow that synchronizes the template every month

zyrafal:
...


zyrafal:
Just testing something

PR #85, opened 2024-08-26
Fix #84: Hide Unauthorized issues

### LINKED ISSUE #84: Hide unauthorized issues
There is a small time window for a malicious user to post a new github issue in https://github.com/ubiquity/devpool-directory with a phishing link. Although such [issue](https://github.com/ubiquity/devpool-directory/blob/61045f468246a6fdde9901a400f0f9c1b348272f/.github/workflows/sync-issues.yml#L51) will be deleted eventually users still can be affected.

We could hide issues created by unauthorized users and keep only the ones created by authorized bots:
1. `76412717: https://github.com/ubiquity`
2. `133917611: https://github.com/ubiquibot`
3. `165700353: https://github.com/UbiquityOS`

ahmedraza1026:
Close newly created or edited issues if they are not from Authorized bots

Resolves #84


Keyrxng (changes requested):
@ahmed-1026 thanks for the PR.

The logic should be embedded into the UI not as a workflow. 

See [`render-github-issues.ts`](https://github.com/ubiquity/work.ubq.fi/blob/development/src/home/rendering/render-github-issues.ts) for one place that might be used to stop the unauthorized issues from rendering. 

The backend effectively deletes any unauthorized issues once every hour, this PR aims to ensure that during that hour nothing can slip through.

ahmedraza1026:
My bad,
working on it.

.github/workflows/hide-issues.yml
Keyrxng, line 1: Please remove this file if not being used

Keyrxng (changes requested):
Is it possible for you to provide some kind of QA in the form of screenshot, video or otherwise?

Typically PRs always come with some kind of visual proof that the spec has been achieved if possible, we can let it slide for this task but for future contributions please keep this in mind

ahmedraza1026:
> Is it possible for you to provide some kind of QA in the form of screenshot, video or otherwise?
> 
> Typically PRs always come with some kind of visual proof that the spec has been achieved if possible, we can let it slide for this task but for future contributions please keep this in mind

image
This code snippet check for the id of creater, if it is in the list then it will proceed, otherwise skip.

ahmedraza1026:
I have updated working, now when issue are rendering it will check if user(creater) of issue is in the list or not.
image
I have request to list valid users in the list in line 17 of src/home/rendering/render-github-issues.ts

Keyrxng:
@ahmed-1026 thanks for the screenshots but what I meant was a screenshot or QA of the actual UI demonstrating the implemented changes. Reviewers can see the changes made to the codebase no problem but seeing the effect the code has requires a reviewer to pull, spin things up locally and find the path needed to trigger it.

It's expected a contributor will have already tested their work to prove it works (to themselves) so grabbing a screenshot at that point is ideal.

---

@0x4007 the implementation here uses the `app_slug` for verifying it's an authorized issue, this seems appropriate although would it be preferred to use the bot userId? We don't have the perk of knowing installs here so it's either or

ahmedraza1026:
> @ahmed-1026 thanks for the screenshots but what I meant was a screenshot or QA of the actual UI demonstrating the implemented changes. Reviewers can see the changes made to the codebase no problem but seeing the effect the code has requires a reviewer to pull, spin things up locally and find the path needed to trigger it.
> 
> It's expected a contributor will have already tested their work to prove it works (to themselves) so grabbing a screenshot at that point is ideal.
> 
> @0x4007 the implementation here uses the `app_slug` for verifying it's an authorized issue, this seems appropriate although would it be preferred to use the bot userId? We don't have the perk of knowing installs here so it's either or

@Keyrxng Thanks for review.
I understand that but for that I have to add an unauthorized issue which I can not.

0x4007:
Always use user id (app id) not app name. 

`236521` `ubiquibot`
`975031` `ubiquity-os`

ahmedraza1026:
I got a conflict that's why I used name instead of id
image
No id is matching with the provided ids.

0x4007:
Strange I assumed that the app id would work in place of the user ID. @Keyrxng any idea? Otherwise we will need to default to app name which doesn't seem right.

Keyrxng:
This was how I did it in my quick fix PR (logic was force-pushed removed so no longer exists in the PR). 

To use the app_id I'm sure we'd need to be able to check the installs, so it's either the bot userId or the app_slug we need to use in this case unless we add support for official install checking


```ts
    if (isBot && userId !== 113181824 && userId !== 76412717 && userId !== 165700353 && userId !== 133917611) {
      continue;
    }
```

Keyrxng:
@ahmed-1026 could you use the `botId` to filter instead of the `app_slug` please and I will merge this after review. So this would be the `user.id` of the bot which posts tasks to the repo.

Please ensure that cypress tests are not broken by any implemented changes also, thanks.

PR #86, opened 2024-08-28
Make regexes handle github links with www

xno-miner:
Some issue links include `www.` which isn't expected by current regexes

This PR makes regexes handle both links with and without `www.`


0x4007:
With the www prefix the github ui does not register a backlink (or "reference") so we made this the behavior on forks to get rid of "references" to all of our issues from developers for the directory. 

These should not be supported. Only the official (authorized) instance should render links without the www prefix. 

I read the context of the original pull and I believe that the correct way to solve the problem is to delete the unauthorized issues posted.

PR #87, opened 2024-08-29
Add dynamic redirectTo handling for GitHub OAuth login

### LINKED ISSUE #82: preview deployment auth
The problem is that our preview deployments redirect us to the production portal and so the auth is not valid and we cannot actually test the deployment.

We use `Supabase.auth.signInWithOAuth` to "Login via GitHub". We can optionally pass a `redirectTo` in the options which may resolve this issue. Either generally redirecting back to the same URL it came from or strictly `work.ubq.fi` and a wildcard to match our preview deployment URLs.

I think it would be safer to restrict.

jordan-ae:
Resolves #82 

This update enhances the GitHub OAuth login process by dynamically setting the `redirectTo` URL based on the current environment.


0x4007:
Didn't work with the test links

PR #89, opened 2024-09-13
chore: create `.ubiquibot-config.yml`

rndquu:
This PR enables gnosis [UUSD](https://gnosisscan.io/address/0xc6ed4f520f6a4e4dc27273509239b7f8a68d2068) payouts for current repository.

As far as I understand, [repo wide](https://github.com/ubiquity/work.ubq.fi/blob/09fba98f84fb9a0e85b8ea3760bbf7a9024249ed/.github/.ubiquibot-config.yml#L4) config overrides [org wide](https://github.com/ubiquity/ubiquibot-config/blob/432ed5bde15c153601f52aa6627af456d3572c81/.github/.ubiquibot-config.yml#L31) one. (@gentlementlegen pls correct me if I'm wrong)

Wiki: https://github.com/ubiquity/ubiquity-dollar/wiki/24.-Smart-Contracts

### DIFF
--- a/.github/.ubiquibot-config.yml
+++ b/.github/.ubiquibot-config.yml
@@ -1,0 +1,92 @@
+plugins:
+  - skipBotEvents: false
+    uses:
+      - plugin: ubiquibot/conversation-rewards@development
+        # we don't skip bot events so conversation rewards triggered by the bot also run
+        with:
+          evmNetworkId: 1
+          evmPrivateEncrypted: "bd5AFnSCO6c5jJyPifpOfr5Zys29RE7SyXkEU3akT13RtGmYDrqGIGuvJQyH53HA5dIba7PL5bXfll0JebmwXYe5gHIXSGX80WuGMDHh0cFfeGjHhmUXe8kkZ1OT2De9qRpqejJcEzdfi-8XNAvP7cQu2Vt-7RNnPw" # https://github.com/ubiquibot/conversation-rewards/pull/111#issuecomment-2348639931
+          erc20RewardToken: "0xb6919Ef2ee4aFC163BC954C5678e2BB570c2D103"
+          incentives:
+            contentEvaluator:
+              enabled: true
+            userExtractor:
+              enabled: true
+              redeemTask: true
+            dataPurge:
+              enabled: true
+            formattingEvaluator:
+              multipliers:
+                - role: [ ISSUE_SPECIFICATION ]
+                  multiplier: 3
+                  rewards:
+                    regex:
+                      "\\b\\w+\\b": 0.1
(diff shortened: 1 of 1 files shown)


gentlementlegen:
The it correct. Currently it is totally overridden. There is a related PR here: https://github.com/ubiquity/ubiquibot-kernel/issues/54

0x4007:
> This PR enables ethereum mainnet [UUSD](https://etherscan.io/address/0xb6919ef2ee4afc163bc954c5678e2bb570c2d103) payouts for current repository.

Let's do the Gnosis Chain UUSD for now.

rndquu:
> > This PR enables ethereum mainnet [UUSD](https://etherscan.io/address/0xb6919ef2ee4afc163bc954c5678e2bb570c2d103) payouts for current repository.
> 
> Let's do the Gnosis Chain UUSD for now.

Updated https://github.com/ubiquity/work.ubq.fi/pull/89/commits/91dc9ecd493bce94bf959094f36548e63565dd39

rndquu:
> The it correct. Currently it is totally overridden. There is a related PR here: [ubiquity/ubiquibot-kernel#54](https://github.com/ubiquity/ubiquibot-kernel/issues/54)

You mean "totally overriden" per plugin, right? So [this](https://github.com/ubiquity/work.ubq.fi/blob/91dc9ecd493bce94bf959094f36548e63565dd39/.github/.ubiquibot-config.yml) config will inherit, let's say, [wallet plugin](https://github.com/ubiquity/ubiquibot-config/blob/432ed5bde15c153601f52aa6627af456d3572c81/.github/.ubiquibot-config.yml#L3) from org wide config.

gentlementlegen:
@rndquu No totally overridden because after changing the shape using the manifest there is no way to know what to keep and what to remove. We should introduce identifiers to know what to keep and what to remove. You can discuss on the spec https://github.com/ubiquity/ubiquibot-kernel/issues/54 if needed. But basically the problem is that right now the configuration is just and array of plugins that can come in any order, and can all have unique chains, hence the problem.

TL;DR for now copy paste the whole configuration and change what you need, this will be fixed in the kernel soon.

0x4007:
I think that means I need to generate another repository specific evm key encrypted, right @rndquu ?

rndquu:
> I think that means I need to generate another repository specific evm key encrypted, right @rndquu ?

No, current private key setup (`PRIVATE_KEY:GITHUB_ORGANIZATION_ID`) will work fine for individual repositories.

Ok, I've copy/pasted the whole config so it seems to be good to merge now.

gentlementlegen (approved):

PR #91, opened 2024-09-19
Debug app name change

### LINKED ISSUE #39: @ubiquity-os Organization no Longer Syncing
I renamed the organization from "Ubiquity-OS" to "ubiquity-os" in the URL. I also renamed the organization "display name" from Ubiquity-OS" to "UbiquityOS"

I noticed now that it no longer 1. displays on https://devpool.directory or 2. syncs to the backend. 

Please research and implement a sustainable fix for if this ever happens again. 

- Example task https://github.com/ubiquity-os/ubiquity-os-plugin-installer/issues/7

work ubq fi_

Keyrxng:
Resolves https://github.com/ubiquity/devpool-directory-bounties/issues/39

`yarn format` hit a few files out of scope but can't be helped.

As far as I can tell the problem is due to the regex used and seems to have resolved the problem. After realizing I thought "this can't be it surely" but as it appears to be, I updated the regex in two locations to be a little more robust.

The backend is updating with issues and they weren't "syncing" with the frontend. I started with debugging the backend first but this seems to be the problem. Would you agree @0x4007?

- https://github.com/ubiquity/devpool-directory/issues/1514

Brave on left, Chrome on right.

image


Keyrxng:
So looking back over things before I sign off for the night, it looks like body is being repeatedly updated back and forth from including `www.` and not. This is the only task that I can see this happening on but it gives me something to look into on the backend specifically. 

I'll add that's it's cool you can see a revision history, this is the first time ever noticing this feature. No way every comment has it wtf how has this gone unnoticed lmao.

image

0x4007:
I don't understand why it's using that www prefix at all. That's not right. The only purpose of that is to prevent "link backs" (references) from the github ui

Also doesn't make any sense that the formatting is so different. Your config must be wrong.

PR #92, opened 2024-09-24
feat: added pwa features

### LINKED ISSUE #90: Progressive Web App
At this conference, there were several instances where I had bad signal, and I generally like to pitch by showing this interface. 

The problem is that it loads the issues after you load the page so it takes a moment and I'm staring at an empty screen essentially.

Turning this into a progressive Web app should allow us to instantly load the cached issues, even when off-line or disconnected

akshaydhayal:
Resolves #90 

This is related to issue #90 . The issues are able to load in local development without the internet.


package-lock.json
0x4007, line 1: Use yarn

Delete this

src/home/fetch-github/fetch-issues-full.ts
0x4007, line 1: Run formatter

static/index.html
```diff
@@ -56,9 +56,23 @@
...
+    <div id="bottom-bar" class="toolbar">
+      <div id="filters-bottom" class="filters-container"></div>
+      <a href="https://dao.ubq.fi/devpool-flow" target="_blank"><button>FAQ</button></a>
+    </div>
+
     <script type="module" src="dist/src/home/home.js"></script>
+    <script>
+      if ('serviceWorker' in navigator) {
+        window.addEventListener('load', () => {          
+          <!-- navigator.serviceWorker.register('static/dist/src/progressive-web-app.js').then(registration => { -->
```
0x4007: This looks like it's trying to be commented out but is still syntax highlighted. Can you clean your code?

0x4007 (changes requested):

src/home/fetch-github/fetch-issues-full.ts
akshaydhayal: I ran the formatter, it is not changing anything to this file?

0x4007, line 1: Run formatter

src/progressive-web-app.ts
```diff
@@ -1,15 +1,55 @@
-self.addEventListener("install", (event: InstallEvent) => {
+const CACHE_NAME = "devpool-directory-cache-v1";
```
0x4007: Maybe it's better to export this and import where needed to increase maintainability.

0x4007:
@rndquu @gentlementlegen can you guys look into continuous deployments? Really inconvenient to QA without it

rndquu:
> @rndquu @gentlementlegen can you guys look into continuous deployments? Really inconvenient to QA without it

I'll check it now

src/progressive-web-app.ts
akshaydhayal: Should I create a constant file for this or use the already github-types file for it?

0x4007: Either I think is fine.

0x4007:
> > @rndquu @gentlementlegen can you guys look into continuous deployments? Really inconvenient to QA without it
> 
> I'll check it now

@rndquu if you were able to identify the problem, and didn't have time to implement a solution - perhaps you can file an issue and one of us can jump on it

src/home/home.ts
```diff
@@ -30,13 +30,44 @@ if (!container) {
...
+    // Update the cache with fresh issues
+    await cache.put("/api/issues", new Response(JSON.stringify(freshIssues)));
+
+    // Request a background sync
+    if ("serviceWorker" in navigator && "SyncManager" in window) {
+      const registration = await navigator.serviceWorker.ready;
+      await registration.sync.register("sync-issues");
+    }
+
+    displayPopupMessage({ modalHeader: "Issues updated", modalBody: "Latest issues have been loaded and cached.", isError: false });
```
0x4007:
```suggestion
```

This seems unnecessary.

src/home/home.ts
```diff
@@ -30,13 +30,44 @@ if (!container) {
...
+async function loadIssues() {
+  try {
+    // First, try to get issues from cache
+    const cache = await caches.open("devpool-directory-cache-v1");
+    const cachedResponse = await cache.match("/api/issues");
+
+    if (cachedResponse) {
+      const cachedIssues = await cachedResponse.json();
+      taskManager.syncTasks(cachedIssues);
+      displayPopupMessage({ modalHeader: "Loaded cached issues", modalBody: "Fetching latest updates...", isError: false });
```
0x4007:
```suggestion
```

Also unnecessary. We don't want to annoy the user with unnecessary popups. These might be more appropriate as console.logs etc.

src/home/home.ts
```diff
@@ -4,7 +4,7 @@ import { initiateDevRelTracking } from "./devrel-tracker";
 import { fetchAndDisplayPreviewsFromCache } from "./fetch-github/fetch-and-display-previews";
 import { fetchIssuesFull } from "./fetch-github/fetch-issues-full";
 import { readyToolbar } from "./ready-toolbar";
-import { renderErrorInModal } from "./rendering/display-popup-modal";
+import { renderErrorInModal, displayPopupMessage } from "./rendering/display-popup-modal";
```
0x4007:
```suggestion
import { renderErrorInModal } from "./rendering/display-popup-modal";
```

0x4007 (commented):
Were you able to test while logged in and authenticated?

0x4007:
image

I got it to load six issues fully but it does not work offline as per the spec 

image

It seems like you're certainly on the right track though.

src/home/fetch-github/fetch-issues-full.ts
0x4007, line 1: Need to run formatting

0x4007 (changes requested):

src/home/fetch-github/fetch-issues-full.ts
akshaydhayal: What do you mean by formatting exactly? If you are referring to the yarn format cmd, I did do that and it was the same code which I committed .

akshaydhayal: If you are referring to the yarn format cmd, I did do that and it was the same code which I committed .

0x4007: Indentation is wrong in your diff. Not sure what to tell you but it looks like you're not using our prettier settings. There is a possibility this was committed without the correct formatting but highly unlikely because we use a template with the same formatter in all of our projects.

akshaydhayal: The formatter is only producing these indentations. If I manually remove these indentation and run the formatter, they are produced again.

See the formatter did checked this file: image

akshaydhayal: Indentations seems correct if you checkout the code, I added try catch blocks, maybe that's why it might be showing in code changes.

akshaydhayal: @0x4007 , what to do for this change?

rndquu (commented):
@akshaydhayal 

The PR works fine. Could you resolve the code conflicts?

One more question:
When running on localhost the UI is opened fine without an internet connection while the latest [development preview](https://2cd78bdc.devpool-directory-ui.pages.dev/) page doesn't work offline. What could be the issue?

src/home/fetch-github/fetch-issues-full.ts
rndquu: I also don't see formatting issues in the [fetch-issues-full.ts](https://github.com/ubiquity/work.ubq.fi/blob/53739ad1591b0b60b6952a8390349738d77c13be/src/home/fetch-github/fetch-issues-full.ts) although the PR diff shows new indentations.

0x4007: Perhaps spaces were changed to tabs.

0x4007:
I think redoing the pull would be easier because a lot of changes were made in the last two weeks.

PR #94, opened 2024-09-26
feat: proposal viewer

### LINKED ISSUE #54: Proposals/Request for Comment View (`rfc.ubq.fi`)
I've opened up a couple of proposals, essentially requests for comment. It could be useful to aggregate them onto a single view for ease-of-access to the DAO. Recent examples:

- https://github.com/ubiquity/.github/issues/111
- https://github.com/ubiquity/.github/issues/110

I imagine that we can find issues without pricing and populate them under the "proposals" view. 

1. Fork this repository, make `rfc.ubq.fi` repository
2. Fork `devpool-directory` repository, make `devpool-rfc` repository 
3. `rfc.ubq.fi` UI should load from `devpool-rfc` "backend"

### A Note on Forks

It seems that we have use cases for similar UI code but for different purposes. Perhaps it makes sense to fork and create new repos (leaderboard.ubq.fi, rfc.ubq.fi.)

In addition, we already have https://github.com/ubiquity/devpool-directory and https://github.com/ubiquity/devpool-directory-private. Ideally we could consolidate everything into a single aggregator repository, and collect even the github issues without pricing. That way we only need to maintain a single repository (instead of potentially three) for our "backend" of issues. 

For now lets just do another fork and then we can worry about consolidating them into a "single backend" in a future project.

### LINKED ISSUE #115: Add support for proposals on `work.ubq.fi`

0x4007:
Resolves https://github.com/ubiquity/.github/issues/115
Resolves https://github.com/ubiquity/work.ubq.fi/issues/54

Click the title "DevPool Directory" and it will instantly switch to the "DevPool Proposals" view


0x4007:
Cypress is unreliable so I'm ignoring it.

0x4007:
It's too bad I can't revert from mobile.

PR #95, opened 2024-09-26
fix: mobile header overflows

0x4007:
Overflowing headers on mobile so had to push a hot fix. 

image


0x4007:
Cache on my phone is not cooperating so it's hard to fully test but on narrow resolutions on my computer it worked as expected. 

Rate limits are more difficult to deal with now with the additional issues. We need to handle this.

PR #99, opened 2024-09-26
Redo Cache System

### LINKED ISSUE #98: Redo Cache System
Now we compile all the issues (proposals and tasks) in a single json [here](https://github.com/ubiquity/devpool-directory/blob/development/devpool-issues.json).

This will solve the following problems:
1. rate limits/requirement for authentication
2. simplify cache

Load this instead of loading everything individually from the GitHub API

patric-obot:
Resolves #98

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@octokit/request-error": "^6.1.0",
     "@octokit/rest": "^20.0.2",
     "@supabase/supabase-js": "^2.39.0",
+    "axios": "^1.7.7",
     "dotenv": "^16.3.1",
     "esbuild-plugin-env": "^1.0.8",
     "marked": "^11.0.0"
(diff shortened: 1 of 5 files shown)


0x4007 (changes requested):
We have private issues that require auth so you can't remove octokit

Also you should just use fetch. There's no reason to use axios in the front end in 2024

0x4007:
Fetch is a browser standard now. It does what you need and you don't need a redundant library to handle network requests

> So means for private issues I shouldn't fetch from the Jason file right?

Right

0x4007 (changes requested):
There's no changes to the cache system

0x4007:
What profile? 

I don't see how else we can solve for this but caching makes it so that we should only request once per org. We only have three orgs so either the cache isn't implemented correctly or you're testing too much with only three orgs.

PR #104, opened 2024-10-03
Enhancement/cached issues

### LINKED ISSUE #98: Redo Cache System

0x4007:
Resolves #98


0x4007:
image


Getting this will have to handle in a new task

src/progressive-web-app.ts
0x4007, line 1: This was pretty nasty but should be fixed in https://github.com/ubiquity/work.ubq.fi/issues/90

PR #106, opened 2024-10-07
feat: optimize avatar fetching from GitHub API and avatar storing as well

### LINKED ISSUE #105: Optimize Avatar Requests
> image
> 
> 
> Getting this will have to handle in a new task
> 

_Originally posted by @0x4007 in https://github.com/ubiquity/work.ubq.fi/issues/104#issuecomment-2392454878_

zugdev:
Resolves #105

This allows only a single request for avatar per organization in `cachedTasks`. It works by allowing a single execution of `fetchAvatar(orgName)` to run per `orgName` while others are kept on wait. On resolve all of them share the same cached blob. This also means that a single instance is setting cache. 

I've also removed the unnecessary avatar fetching from DB in `fetch-and-display-preview.ts`, since it's `fetch-avatar.ts`'s single responsibility to do so. By keeping all avatar-related logic within `fetchAvatar`, we ensure a single source of truth for avatar fetching, which simplifies code too.

### DIFF
--- a/src/home/fetch-github/fetch-and-display-previews.ts
+++ b/src/home/fetch-github/fetch-and-display-previews.ts
@@ -1,5 +1,4 @@
 import { getGitHubAccessToken } from "../getters/get-github-access-token";
-import { getImageFromCache } from "../getters/get-indexed-db";
 import { getLocalStore } from "../getters/get-local-store";
 import { GITHUB_TASKS_STORAGE_KEY, GitHubIssue, TaskStorageItems } from "../github-types";
 import { taskManager } from "../home";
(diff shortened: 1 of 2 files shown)


zugdev:
@0x4007 I've let the `console.log` statements in the code for you to compile, run and check browser console. An easy way to test all fetching method works is commenting each step as you test. A pending orgName fetch will try to get the avatar in the following order:

1. Try to get from DB and set cache
2. Try to get from GitHub using organization search and set cache
3. Try to get from GitHub using user search and set cache

To ensure 2 works well, comment the first step. Similarly if you want to ensure 3 works well, comment the second step.

You can see how a single "thread" caches the avatar for an org below.

image

PR #107, opened 2024-10-07
feat: enable github auth in local and preview environment

### LINKED ISSUE #82: preview deployment auth

zugdev:
Resolves #82

### DIFF
--- a/src/home/rendering/render-github-login-button.ts
+++ b/src/home/rendering/render-github-login-button.ts
@@ -29,25 +29,12 @@ export async function checkSupabaseSession() {
 }
 
 async function gitHubLoginButtonHandler(scopes = "public_repo read:org") {
-  let redirectTo = "https://work.ubq.fi";  // Default to production URL
-
-  // If we are in an authorized preview environment, auth should redirect back to it
-  if (window.location.hostname === "localhost" && window.location.port === "8080") {
-    console.log("Localhost detected, redirecting to localhost:8080");
-    redirectTo = "http://localhost:8080";
-  } else if (window.location.hostname === "devpool.directory") {
-    console.log("Devpool directory detected, redirecting to devpool.directory");
-    redirectTo = "https://devpool.directory";
-  } else if (window.location.hostname === "devpool-directory-ui.pages.dev") {
-    console.log("Devpool directory UI detected, redirecting to devpool-directory-ui.pages.dev");
-    redirectTo = "https://devpool-directory-ui.pages.dev";
-  }
-
+  let redirectTo = window.location.href;
   const { error } = await supabase.auth.signInWithOAuth({
     provider: "github",
     options: {
       scopes,
-      redirectTo,
+      redirectTo
     },
   });
   if (error) {
--- a/src/home/rendering/render-github-login-button.ts
+++ b/src/home/rendering/render-github-login-button.ts
@@ -33,10 +33,13 @@ async function gitHubLoginButtonHandler(scopes = "public_repo read:org") {
 
   // If we are in an authorized preview environment, auth should redirect back to it
   if (window.location.hostname === "localhost" && window.location.port === "8080") {
+    console.log("Localhost detected, redirecting to localhost:8080");
     redirectTo = "http://localhost:8080";
   } else if (window.location.hostname === "devpool.directory") {
+    console.log("Devpool directory detected, redirecting to devpool.directory");
     redirectTo = "https://devpool.directory";
   } else if (window.location.hostname === "devpool-directory-ui.pages.dev") {
+    console.log("Devpool directory UI detected, redirecting to devpool-directory-ui.pages.dev");
     redirectTo = "https://devpool-directory-ui.pages.dev";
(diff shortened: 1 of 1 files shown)


src/home/rendering/render-github-login-button.ts
```diff
@@ -29,10 +29,22 @@ export async function checkSupabaseSession() {
 }
 
 async function gitHubLoginButtonHandler(scopes = "public_repo read:org") {
+  const previewDomainPattern = /(localhost:8080|\.pages\.dev)$/;
+
+  let redirectTo = "https://work.ubq.fi";  // Default to production URL
```
0x4007: There is also Devpool.directory

0x4007 (changes requested):

src/home/rendering/render-github-login-button.ts
zugdev: solved

0x4007:
Doesn't work

zugdev:
@0x4007 Can you please try again with console open? See if a msg pops up and redirect is correct.

rndquu:
@zugdev The issue is that cloudflare [preview deployment](https://github.com/ubiquity/work.ubq.fi/pull/107#issuecomment-2395824462) URLs redirect to `work.ubq.fi` after github oauth instead of the preview URL.

Example:
1. Open https://11352a0e.devpool-directory-ui.pages.dev/
2. Login with github
3. You'll be redirected to `work.ubq.fi` while the expected redirect URL is https://11352a0e.devpool-directory-ui.pages.dev/


Can't we simply use:
```js
  const { error } = await supabase.auth.signInWithOAuth({
    provider: "github",
    options: {
      scopes,
      redirectTo: window.location.origin,
    },
  });
```

zugdev:
> Example
>https://11352a0e.devpool-directory-ui.pages.dev/

So the problem is domain is different from what I expected! Can use wildcard for whatever is that beginning part of the domain.

> 
> Can't we simply use:
> ```js
>   const { error } = await supabase.auth.signInWithOAuth({
>     provider: "github",
>     options: {
>       scopes,
>       redirectTo: window.location.origin,
>     },
>   });
> ```
> 

According to [this discussion](https://github.com/supabase/supabase/discussions/6088#discussioncomment-2468175) this approach doesn't work, though it's very old. I do have a solution in mind will push in a few hours.

zugdev:
It does not work unfortunately, I made a simple deployment to test it: https://supabase-test-laqua7nr8-zugs-projects-e51b52e0.vercel.app

zugdev:
@rndquu @0x4007 According to [docs](https://supabase.com/docs/guides/auth/redirect-urls#use-wildcards-in-redirect-urls), you guys need to go into Supabase dashboard allow list and add all URLs you want to be able to redirect to. I've set up a test project to check it out and it should look something like this for you:

image

image

Use strict URL for the prod URL and use wildcards for URLs where a commit/deploy ID can come in some part of the domain, as I did in: `https://**.devpool-directory-ui.pages.dev`

The link to this URL config page is https://supabase.com/dashboard/project/{project_id}/auth/url-configuration just change `project_id` to the appropriate ID such as `zjypueniamnpwwwgywzs`.

zugdev:
Because of the auth token the allowed redirect list should actually be:

`https://**.devpool-directory-ui.pages.dev/**` with the wildcard at the end. It is very important.

I think the allowed list would be close to:

`https://**.devpool-directory-ui.pages.dev/**` 
`https://devpool-directory-ui.pages.dev/**` 
`https://devpool.directory/**` 
`http://localhost:8080/**`

Set the default site URL to be prod so it can be strict and not allow wildcards. I think it already is, based on current default redirect.

zugdev:
After setting that using `window.location.href` should work! You can hypothetically try it out with my demo at:

1. https://supabase-test-beta.vercel.app
2. [supabase-test-git-master-zugs-projects-e51b52e0.vercel.app](https://supabase-test-git-master-zugs-projects-e51b52e0.vercel.app/)
3. [supabase-test-d7wua39up-zugs-projects-e51b52e0.vercel.app](https://supabase-test-d7wua39up-zugs-projects-e51b52e0.vercel.app/)

rndquu:
@zugdev Added

Screenshot 2024-10-07 at 16 50 55

rndquu:
@zugdev Preview deployment [failed](https://github.com/ubiquity/work.ubq.fi/actions/runs/11216992647/job/31177521722) due to rate limits, let's wait some time

rndquu (approved):
Works fine

0x4007 (approved):

PR #114, opened 2024-10-08
feat: optimize code for fetching, caching and task handling

### LINKED ISSUE #108: Remove Redundant Calls and Optimize Caching
While working through some issues I noticed there are some caching optimizations pending, redundant calls being made and load time could improve. A clear example was that for each issue the organization's avatar was fetched, meaning that for 144 issues within 3 repos, where 3 calls could have been made we were making 144. This was solved at #106. 

Some code parts are also not respecting single responsibility principle, there are tries to set cache outside of a fetch function. This leads to resource waste and subsequently worse performance. 

For example, the following O(n) function where n is issues array length is being called 4 different times in different places:

```
export function applyAvatarsToIssues() {
  const container = taskManager.getContainer();
  const issueElements = Array.from(container.querySelectorAll(".issue-element-inner"));

  issueElements.forEach((issueElement) => {
    const orgName = issueElement.querySelector(".organization-name")?.textContent;
    if (orgName) {
      const avatarUrl = organizationImageCache.get(orgName);
      if (avatarUrl) {
        const avatarImg = issueElement.querySelector("img");
        if (avatarImg) {
          avatarImg.src = URL.createObjectURL(avatarUrl);
        }
      }
    }
  });
}
```

image

The code base could benefit from general optimizations and a principle-based approach. I believe a 2-3 hour thorough review should be enough to identify, test and fix some of the problems.

zugdev:
Resolves #108

You can see some of my thoughts as comments at [this commit](https://github.com/ubiquity/work.ubq.fi/commit/4d902f41f4ed7fb52b1dbeaf3025a6a13bf4f01d) files. 

The approach here is to:

1. Implement single responsibility amongst functions, i.e:. fetch should fetch, handler should handle...

2. Structure routine to avoid redundant calls. `displayGitHubIssues` is called multiple times from multiple sources and shouldn't be, and so is `fetchAvatars` and `fetchAndDisplayIssuesFromNetwork` - whose name no longer makes sense.


src/home/fetch-github/preview-to-full-mapping.ts
```diff
@@ -1,5 +1,6 @@
 import { GitHubIssue } from "../github-types";
 
+// this is not used
```
0x4007: There are some comments that seem like action items. Why not just implement? This pull doesn't seem ready.

0x4007 (changes requested):

src/home/fetch-github/preview-to-full-mapping.ts
zugdev: it wasn't ready, my bad opened wrong type of PR

zugdev:
I've removed the login check that resets cache because once user authenticates the page is reloaded, which means that tasks will be synced anyway.

0x4007 (approved):
Code looks fine and QA seems fine from mobile.

PR #123, opened 2024-10-12
feat: implement new tracker

### LINKED ISSUE #74: Reimplement Tracking Code

koya0:
Resolves #74


koya0:
the code I used for the worker was this js file:

```js
addEventListener('fetch', event => {
  event.respondWith(handleRequest(event.request));
});

const corsHeaders = {
  "Access-Control-Allow-Origin": "*",
};

async function handleRequest(request) {
  const url = new URL(request.url);

  if (url.pathname === '/set') {
    const key = url.searchParams.get('key');
    const value = url.searchParams.get('value');
    
    if(key && value) {
      await userToReferral.put(key, value);

      return new Response(`Key '${key}' added with value '${value}'`, { headers: corsHeaders, status: 200 });
    } else {
      return new Response('Missing key or value in query params', { headers: corsHeaders, status: 400 });
    }
  }

  if (url.pathname === '/get') {
    const key = url.searchParams.get('key');

    if (key) {
      const value = await userToReferral.get(key);

      return value ? 
        new Response(`Value for '${key}': ${value}`, { headers: corsHeaders, status: 200 }) : 
        new Response(`No value found for '${key}'`, { headers: corsHeaders, status: 404 });
    } else {
      return new Response('Missing key in query params', { headers: corsHeaders, status: 400 });
    }
  }

  if (url.pathname == '/list') {
    const keys = await userToReferral.list(); // get at max 1000 keys
    const keyValuePairs = {};

    for (const key of keys.keys) {
      const value = await userToReferral.get(key.name);
      keyValuePairs[key.name] = value; 
    }

    return new Response(JSON.stringify(keyValuePairs, null, 2), {
      headers: {...corsHeaders, 'Content-Type': 'application/json' },
    });
  }
  return new Response('Not Found', { headers: corsHeaders, status: 404 });
}
```
The name of the KV namespace I used was "userToReferral" and I had to use the corsHeaders on the responses because they were getting blocked by CORS policy, when sent to the localhost.

koya0:
I also made a quick demo of the tracker working

https://github.com/user-attachments/assets/864f34d3-44a1-4076-8f32-50a7fcbcf658

zugdev:
> the code I used for the worker was this js file:

I believe it's best if you commit this, perhaps `cloudflare-tracking-worker.js`.

0x4007:
Check the beta branch of pay.ubq.fi repo to see how we committed cloudflare workers along with the UI code. It automatically deploys with our CI as well. 

Maybe @gentlementlegen @whilefoo and @EresDev can explain more on it

whilefoo:
@koya0 you can put worker code as a [page function](https://developers.cloudflare.com/pages/functions/get-started/). If you put the code in `/functions/tracker.ts` you can call it with `https://worker-url.dev/tracker`

src/home/tracking/cloudflare-tracking-worker.js
whilefoo, line 1: this is not secure, you need to verify the access token either through Github or Supabase Auth

whilefoo (changes requested):

src/home/tracking/cloudflare-tracking-worker.js
koya0: You want to make sure that we handle the request, only if the user is logged in with the GitHub account?

functions/helpers.js
```diff
@@ -0,0 +1,3 @@
+export const corsHeaders = {
+	"Access-Control-Allow-Origin": "*",
```
0x4007: Be sure to run `yarn format`

0x4007 (changes requested):
Commit TypeScript not JavaScript and run formatting

src/home/tracking/cloudflare-tracking-worker.js
whilefoo: Yes

0x4007 (commented):
Code looks good. How do we test?

functions/tracker.ts
```diff
@@ -37,11 +38,20 @@ export async function onRequest(ctx: Context): Promise<Response> {
   }
 }
 
-async function handleSet(url: URL, env: Env): Promise<Response> {
+async function handleSet(url: URL, env: Env, request: CustomRequest): Promise<Response> {
   const key = url.searchParams.get("key");
   const value = url.searchParams.get("value");
```
whilefoo: these variables should have names that describe what they are

src/home/devrel-tracker.ts
```diff
@@ -18,8 +20,14 @@ export async function trackDevRelReferral(devGitHubId: number) {
   if (devRelCode && devRelCode != "done") {
     const url = `${WORKER_URL}/tracker?key=${encodeURIComponent(devGitHubId)}&value=${encodeURIComponent(devRelCode)}`;
```
whilefoo: key is not needed anymore because the user ID is in the token. I think it's best to send the referral code and token in the body because it's a POST request

functions/validators.ts
```diff
@@ -0,0 +1,22 @@
+import { CustomRequest } from "./types";
+import { OAuthToken } from "../src/home/getters/get-github-access-token";
+
+export async function validatePOST(url: URL, request: CustomRequest): Promise<boolean> {
```
whilefoo:
```suggestion
export async function validateGithubToken(token: string): Promise<boolean> {
```
this function should call Github API to verify that the token is actually valid
also I think the token is JWT in string format so you need a JWT library to decode it

functions/types.ts
```diff
@@ -1,7 +1,10 @@
...
 
 export interface Env {
   userToReferral: KVNamespace;
 }
 
-export type Context = EventContext<Env, string, Record<string, string>>;
\ No newline at end of file
+export type Context = EventContext<Env, string, Record<string, string>>;
+
+export type CustomRequest = Request<unknown, IncomingRequestCfProperties<unknown>>;
```
whilefoo: are you sure this is needed? I've never needed to use it, usually just `Request`

whilefoo (changes requested):

functions/validators.ts
koya0: I guess you didn't see the most recent commit I made for validatePOST. I changed a few things, and now, I'm checking if the authToken is valid with Octokit, you can check the commits: [here](https://github.com/ubiquity/work.ubq.fi/pull/123/commits/4667f1245aea9c72b5e6e299af9d029d8eb18162) and [here](https://github.com/ubiquity/work.ubq.fi/pull/123/commits/c59954d2cf4ae014ff9d1ed89a07834cf595304f)

functions/types.ts
koya0: the type of the request I get here: `const { request, env } = ctx;` is the `Request<unknown, IncomingRequestCfProperties<unknown>>`, I tried to type cast the variable `request` to `Request`, but it was not successful, I think this strange request is specific to Cloudflare Workers.

koya0:
> Code looks good. How do we test?

1. Clone [my repo](https://github.com/koya0/work.ubq.fi)
2. Create a KV Namespace with ```wrangler kv:namespace create "name of your kv"```
After running this, you'll get the id of the KV, store that id

Now you have 2 options:
3. Change the KV id of the wrangler.toml by yours
4. Deploy a new page with ```wrangler pages deploy ./static --project-name test-tracker```
5. Set up .env WORKER_URL with the url of your page
or
3. Send me the id of your KV
4. I will deploy my page with your kv
5. I will send you my page url for you to set up the .env WORKER_URL

6. Run yarn and login with a ref code on local host
7. You can check if a pair was added to the your kv in the [cloudflare dashboard](https://dash.cloudflare.com/)

src/home/devrel-tracker.ts
```diff
@@ -13,12 +13,12 @@ export function initiateDevRelTracking() {
...
 }
 
-export async function trackDevRelReferral(devGitHubId: number) {
+export async function trackDevRelReferral() {
   const devRelCode = localStorage.getItem("devRel");
 
   // key: user_id (devGitHubId), value: referral_id (devRelCode)
   if (devRelCode && devRelCode != "done") {
-    const url = `${WORKER_URL}/tracker?key=${encodeURIComponent(devGitHubId)}&value=${encodeURIComponent(devRelCode)}`;
+    const url = `${WORKER_URL}/tracker`;
```
whilefoo: you don't need worker URL because it is the same as the page URL so you can just use `/tracker`

functions/validators.ts
```diff
@@ -0,0 +1,43 @@
...
+      console.error("Missing provider token");
+      return { isValid: false };
+    }
+
+    const octokit = new Octokit({ auth: providerToken });
+
+    try {
+      await octokit.request("GET /user");
+
+      const githubUserId = authToken.user?.user_metadata?.provider_id;
```
whilefoo: You should send Github auth token instead of the whole Supabase auth object which can be manipulated with and you can get the user ID from the octokit request.
Also why are there 2 try/catch blocks?

whilefoo (changes requested):

functions/types.ts
whilefoo: I have just tried locally with `Request` from `@cloudflare/workers-types` and it works normally

functions/validators.ts
koya0: I was using the first one to handle the json parsing, but I will remove it

functions/types.ts
koya0: Oh sorry, probably when I tried to use `Request` the import got it from another lib, and it didnt work, so I end up using that other one, I will update to use the `Request` from `@cloudflare/workers-types`

koya0:
> Code looks good. How do we test?

you can login in the [page I deployed](work-ubq-fi-d8j.pages.dev) with a ?ref= in the url, after the auth, the pair {githubId, referralCode} will be added to my kv, then I can send a screenshot from it, and you can check if the referralCode is correct

or

you can clone [my repo](https://github.com/koya0/work.ubq.fi/tree/development), create your kv namespace, deploy a new page, changing the kv id from wrangler.toml to yours, and then check if your kv gets the pair

I wasn't able to test the tracker with the usual supabase url, because after the auth I was being redirected to `devpool.directory`, and the tracker was not working, so in my page, I'm using a supabase @zugdev set. [More details](https://github.com/ubiquity/work.ubq.fi/issues/74#issuecomment-2424266004)

zugdev:
You should remove the `?ref` parameter from the URL after authentication. You can do that with:

```
const newURL = new URL(window.location.href);
newURL.searchParams.delete("ref");
window.history.pushState({}, "", newURL.toString());
```

zugdev:
I have tried it, please send screenshot proving my ref was captured. I have recorded video with time in it.

koya0:
> I have tried it, please send screenshot proving my ref was captured. I have recorded video with time in it.

got it!
image

zugdev:
Looks good!

[Screencast from 19-10-2024 19:30:46.webm](https://github.com/user-attachments/assets/2f140820-c960-4d6a-91e2-50ba686cc37b)

zugdev:
In practice ref will be GitHub IDs too, right?

koya0:
@0x4007 @whilefoo have you already tested? What should be done next? Just to know if you are waiting for me 😅

src/home/devrel-tracker.ts
```diff
@@ -1,22 +1,44 @@
...
   const oldDevRelCode = localStorage.getItem("devRel");
   if (!oldDevRelCode) {
     const urlParams = new URLSearchParams(window.location.search);
-    const devRelCode = urlParams.get("devRel");
+    const devRelCode = urlParams.get("ref");
     if (devRelCode) {
       localStorage.setItem("devRel", devRelCode);
     }
   }
 }
```
0x4007: I think we should remove support for the old code. We don't need it.

src/home/devrel-tracker.ts
```diff
@@ -1,22 +1,44 @@
...
 
-export function trackDevRelReferral(devGithub: string) {
+export async function trackDevRelReferral() {
   const devRelCode = localStorage.getItem("devRel");
```
0x4007: Remove the old support here too.

src/home/getters/get-github-referrals.ts
```diff
@@ -0,0 +1,34 @@
...
+    // No referral id found for devGitHubId
+    return null;
+  } else {
+    console.error(`Failed to get key: '${devGitHubId}'. Status: ${response.status}`);
+    return null;
+  }
+}
+
+export async function getListOfReferrals(): Promise<Record<string, string | null> | null> {
+  const url = "/tracker";
```
0x4007: Using terms like "tracker" will definitely get blocked by ad blockers.

wrangler.toml
0x4007, line 1: These seem like they should be secrets?

whilefoo (commented):
@0x4007 how do we want to handle multiple referrals? 
I think we should only count the first referral. If you come to the page with new referral code we don't save it

zugdev:
> @0x4007 how do we want to handle multiple referrals? I think we should only count the first referral. If you come to the page with new referral code we don't save it

Yes, current solution appears to be a mapping so this would be accounted for. Is that correct, @koya0 ?

koya0:
> > @0x4007 how do we want to handle multiple referrals? I think we should only count the first referral. If you come to the page with new referral code we don't save it
> 
> Yes, current solution appears to be a mapping so this would be accounted for. Is that correct, @koya0 ?
> 
https://github.com/koya0/work.ubq.fi/blob/development/src%2Fhome%2Fregister-referral.ts#L8-L12

Yes, I only insert a pair to the kv, if there isn't a value associated to the key, I check this with `getReferralFromUser`

koya0:
> > @0x4007 how do we want to handle multiple referrals? I think we should only count the first referral. If you come to the page with new referral code we don't save it
> 
> Yes, current solution appears to be a mapping so this would be accounted for. Is that correct, @koya0 ?

https://github.com/ubiquity/work.ubq.fi/blob/6f5fa3ee34d66c2dcc7b723a584c91df1403f358/src/home/register-referral.ts#L8-L12

Yes, before adding a pair `{gitHubUserId, referralCode}` to the kv, I check if there is already a value associated with the key with `getReferralFromUser`, if the function returns `null`, then I add the pair, so only the first referral code counts.

whilefoo:
> Yes, before adding a pair `{gitHubUserId, referralCode}` to the kv, I check if there is already a value associated with the key with `getReferralFromUser`, if the function returns `null`, then I add the pair, so only the first referral code counts.

That's fine but it'd make more sense for that check to be on the backend/functions

whilefoo (approved):

whilefoo:
@koya0 please resolve the conflicts and change start command to wrangler

whilefoo (changes requested):
sorry to keep bothering you, just a small change: you should store ref to the local storage right away and then use it after the user signs it like [this](https://github.com/ubiquity/work.ubq.fi/blob/cb5e4ad8b3e71768074968e9f64a1f2c8812dd80/src/home/devrel-tracker.ts#L1) function that was used before. this is because the user can leave the page and come back again but referral should be still counted

koya0:
> sorry to keep bothering you, just a small change: you should store ref to the local storage right away and then use it after the user signs it like [this](https://github.com/ubiquity/work.ubq.fi/blob/cb5e4ad8b3e71768074968e9f64a1f2c8812dd80/src/home/devrel-tracker.ts#L1) function that was used before. this is because the user can leave the page and come back again but referral should be still counted

That's fine! But, I thought @0x4007 wanted me to remove that, when he asked me to "remove the support for the old code" [here](https://github.com/ubiquity/work.ubq.fi/pull/123#discussion_r1812055622) and [here](https://github.com/ubiquity/work.ubq.fi/pull/123#discussion_r1812056439), but maybe I didn't understand what he really wanted. I will fix it!

koya0:
> @koya0 please resolve the conflicts and change start command to wrangler

I fixed the conflicts, but I'm not sure what start command you are referring to, the one in `package.json`?

whilefoo:
> That's fine! But, I thought @0x4007 wanted me to remove that, when he asked me to "remove the support for the old code" [here](https://github.com/ubiquity/work.ubq.fi/pull/123#discussion_r1812055622) and [here](https://github.com/ubiquity/work.ubq.fi/pull/123#discussion_r1812056439), but maybe I didn't understand what he really wanted. I will fix it!

I think he meant the google analytics part of the code



> I fixed the conflicts, but I'm not sure what start command you are referring to, the one in `package.json`?

Yes the `start` command in `package.json` should be changed to `wrangler pages dev --port 8080` (it needs to be 8080 because its whitelisted in the ubiquity supabase redirect list for testing locally). Also remove `@cloudflare/wrangler` because its deprecated and new package name is just `wrangler`

koya0:
image

I got a new pair in the kv, was it you @whilefoo ?

whilefoo:
> I got a new pair in the kv, was it you @whilefoo ?

Yes

whilefoo:
I realized we have two lock files, yarn is used by cypress so that's why its failing. Can you run `yarn install`, remove bun lock file and commit

koya0:
> I realized we have two lock files, yarn is used by cypress so that's why its failing. Can you run `yarn install`, remove bun lock file and commit

when I ran yarn install, the file didn't change

whilefoo:
Are you sure? When I run it, it changes

koya0:
> Are you sure? When I run it, it changes

I'm running with yarn 3.6.3, what is your version?

zugdev:
> Are you sure? When I run it, it changes

yarn might change structure only. it's very unpredictable, when I run in Ubuntu/Arch/Windows it generates completely different files from same package.json

Keyrxng:
The org uses `yarn@v1.2` in 95% of repos, there are a select few which use bun or a particular version of yarn, that's usually `4.22` or `4.2.2` maybe afaik. The `yarn.lock` gives hints as to what version generated the file.

I've advocated to the use `packageManager` field of the `package.json` to avoid this very scenario amongst other reasons.

0x4007:
I use yarn 1.22.21 and don't have these problems.

zugdev:
that's the problem with package managers, first you create npm, than with multiple versions people need to fix a version in each project, people have multiple projects so they need to run an npm version manager lol. soon you will need a manager for nvm which manages node versions

0x4007:
We use node 20.10.0 across all our CI

I suppose it might make sense to enforce that in the package.json as well.

koya0:
> Are you sure? When I run it, it changes

@whilefoo I changed my yarn to v1.22.22 and updated the yarn.lock, maybe now it won't change

whilefoo (approved):
Tests are not passing but they also don't pass on the main branch so this is not a problem caused by your code, I think we're good to merge

0x4007 (commented):
Looks like you now have two lock files. Is that correct? Actually it's a bit hard for me to tell for sure from mobile

0x4007 (approved):

koya0:
I was having problems deploying the page with workers, because the version of yarn they use is 3.6.3, and it was getting conflicts, so I added a variable in wrangler to specify the yarn version worker will use.

0x4007:
Where's the data stored then? How to get this in our database? @whilefoo perhaps you can tie up odds and ends if any?

koya0:
> Where's the data stored then? How to get this in our database? @whilefoo perhaps you can tie up odds and ends if any?

It's stored in my KV Namespace, you need to create a KV and then change the KV id in wrangler.toml to yours.

0x4007:
> > Where's the data stored then? How to get this in our database? @whilefoo perhaps you can tie up odds and ends if any?
> 
> 
> 
> It's stored in my KV Namespace, you need to create a KV and then change the KV id in wrangler.toml to yours. 

@gentlementlegen i think this might be easiest for you to set up given your recent kv work?

gentlementlegen:
All of the KV setup is automated via the Action deployment script, these variables should be described there. [This](https://github.com/ubiquity/work.ubq.fi/pull/123/files#diff-b1b9719b6eae4103d520f1e902420f5073b5e18a5a47aa73feed71a037557c98R9) should not have been commited as such. Example from the plugin: https://github.com/ubiquity-os/plugin-template/blob/development/.github/workflows/worker-deploy.yml#L34

By hardcoding the bindings if we change org / try to deploy on our own, it will break.

rndquu:
Is it working right now on production? I've tried it with `https://work.ubq.fi/?ref=test` but the KV store is empty in the `Ubiquity DAO Workers` project in cloudflare dashboard.

0x4007:
Somebody with cloudflare access @gentlementlegen @rndquu please fix

gentlementlegen:
This should not be fixed within cloudflare because the next deployment will erase the KV most likely, at least that's the behavior I had with earlier versions of wrangler. KV and secrets should be deployed from the actions like we do for plugins, refer to [this comment](https://github.com/ubiquity/work.ubq.fi/pull/123#issuecomment-2440249980).

0x4007:
koya0 doesn't have access to our cloudflare and can't fix this problem.

gentlementlegen:
We do not need Cloudflare to fix this problem but a proper deployment Action script.

0x4007:
Then I suppose a task should be created because I wouldn't be able to implement a fix based on the context I have.

0x4007:
@gentlementlegen

PR #126, opened 2024-10-15
feat: notifications

### LINKED ISSUE #117: Notifications View/Filters
## Remarks

I'm still debating if we should consolidate all "list like" UIs on here, but I'm optimistic it could make sense to (like leaderboards as well.) If we include all these UIs here, it may make sense to rebrand to os.ubq.fi. Also the title bar from DevPool Directory to UbiquityOS: 
- Directory
- Proposals
- Notifications
- XP Leaderboard 

## Overview

As activity continues to pick up and my notifications become slower to get through, it would be useful to make a simple ui that allows for more intelligent notification filtering. 

### Prioritization Sorting Algorithm

- We need to detect the priority level of the task the notification is associated with.
   - If the notification is not associated with a task, consider not showing it on this view at all. 
   - If the notification is a pull request, it must be linked to a task with a priority label.
- Reverse chronological order, meaning that neglected/old notifications should appear higher than new ones. 
- Top of list is highest priority, then work your way down to low priority. 

### Blocking Tasks

If it thinks without my input the task is blocked, it's a higher priority notification.
- Linkbacks: if the notification is associated with a task with many linkbacks then it should be considered higher priority, it usually means its blocking other pulls/tasks

Perhaps we could have an AI layer on top one day to summarize and/or reorder notifications from most important to least.

### Remarks
- Always exclude draft pulls.

0x4007:
Resolves #117


0x4007:
Winding down for the day but got a jumpstart on this.

Mixed feelings on combining this quite different logic in here. Now we have tasks, proposals, and notifications view. The first two are quite similar, as they are both GitHub Issues. Notifications have quite a different shape. 

localhost_8080_ (1)

There's a slim chance I might not merge this in but I do want to see how far I can take this and see if its useful.

0x4007:
Yeah it adds a ton of complexity I think its better to do a separate repo actually. 

- This will benefit greatly from the devpool-directory backend to aggregate every issue in the network, regardless of its assigned state. 
- Then pull all of them in one shot and do local comparisons instead of requesting all of the assigned issue/pull details individually per notification (for the sorting algorithm to consider the priority level)

zugdev:
We should include PRs in the notifications too.

PR #129, opened 2024-10-19
feat: add issues in URL and handle history state

### LINKED ISSUE #118: Deep Links
- We should push and pop browser history when opening issues. 
- We need to account for proposals and for tasks. 
- We need to automatically open the task/proposal when the page is loaded with the url representing it. Maybe a # in the URL with the issue ID will be sufficient.

zugdev:
Resolves #118

### DIFF
--- a/src/home/rendering/render-github-issues.ts
+++ b/src/home/rendering/render-github-issues.ts
@@ -5,6 +5,7 @@ import { taskManager } from "../home";
 import { renderErrorInModal } from "./display-popup-modal";
 import { closeModal, modal, modalBodyInner, titleAnchor, titleHeader } from "./render-preview-modal";
 import { setupKeyboardNavigation } from "./setup-keyboard-navigation";
+import { isProposalOnlyViewer } from "../fetch-github/fetch-and-display-previews";
 
 export function renderGitHubIssues(tasks: GitHubIssue[]) {
   const container = taskManager.getContainer();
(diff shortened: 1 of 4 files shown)


src/home/rendering/render-github-issues.ts
```diff
@@ -149,6 +149,40 @@ export function viewIssueDetails(full: GitHubIssue) {
   modal.classList.add("active");
   modal.classList.remove("error");
   document.body.classList.add("preview-active");
+
+  updateURLWithIssueID(full.id);
```
0x4007: Strict pascal case lint shouldnt let this work

0x4007 (commented):
Try running linter. Commit hooks shouldnt have let you commit these

0x4007:
Just tested. Looks like it is almost perfect. I think ideally the app mode should change to proposals when viewing a proposal though. 

https://4a76efd5.devpool-directory-ui.pages.dev/?issue=2499194812

This is a proposal but when I X out of the modal, its in the tasks view. Any good solution for this?

zugdev:
Yeah that's easy will check `isProposalsViewer`, should be effective enough.

zugdev:
Should be good to go now. I have added a proposal parameter to the URL if we are in proposal view. Also it changes the toggle from Directory to Proposals based on URL.

0x4007 (approved):

PR #130, opened 2024-10-19
feat: fix keyboard navigation buttons overlay

### LINKED ISSUE #127: keyboard navigation buttons over commit id in certain page widths
As you can see in the screenshot (at the bottom-right corner), for some tablet-like widths, the navigation buttons are overlapping with the commit id. On mobile view the commit id is hidden, so just making sure the size triggers are the same to show the buttons and hide the commit id should fix it.

image

zugdev:
Resolves #127

### DIFF
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -373,7 +373,7 @@
   .full {
     display: unset !important;
   }
-  @media screen and (max-width: 560px) {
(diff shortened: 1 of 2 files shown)


0x4007 (approved):
The spec stated that it should be moved to the left but I suppose hiding works fine as well.

PR #131, opened 2024-10-19
feat: offline working progressive webapp 

zugdev:
Resolves #90

Currently working offline by caching issues. When we load we get issues from cache and start trying to fetch from json. If it's successful we check if there are new issues, if so we cache them and re-display issues.

No network demo:

[noNetwork.webm](https://github.com/user-attachments/assets/8099ce8e-dca7-4df9-8ec2-d13f9a3e4578)

Cached issues are in sync:

[alreadyCached.webm](https://github.com/user-attachments/assets/a024da43-5732-4225-850f-d9069d3c5893)

Cached issues not in sync:

[fetchAndCache.webm](https://github.com/user-attachments/assets/38493c07-15cb-4098-b6db-8cb88ceeaa23)

PR #136, opened 2024-10-23
fix: prevent re-render on issue fetching and improve PWA

zugdev:
#131 introduced issue caching to improve performance on poor networks. Now we render issues from cache and fetch in parallel. If cached issues aren't up-to-date we cache fetched issues and render them. However, this leads to a replay in the appearing issues animation. This PR introduces an option to skip this animation, handling both the case where a stale issue is removed and a new issue is introduced. Here is a demo:

[fix-rerender-demo.webm](https://github.com/user-attachments/assets/78398eef-4ba8-4da9-bb86-3a8e4081e6c8)


zugdev:
I've also changed the strategy in the PWA. We should default to Network and if offline load from cache. Defaulting to cache is horrible DX because you constantly need to clear cache in order to see update in code. This needs review and merge ASAP @0x4007 

My vercel deployment: https://work-ubq-fi-git-fix-rerender-zugs-projects-e51b52e0.vercel.app

0x4007:
Vercel is private but requested access. Why use vercel and not our continuous deploy?

static/progressive-web-app.js
```diff
@@ -43,29 +42,21 @@ self.addEventListener("activate", (event) => {
...
+      .catch(() => {
+        return caches.match(event.request).then((cachedResponse) => {
+          if (cachedResponse) {
+            return cachedResponse;
+          } else {
+            return null;
+          }
         });
-    })
+      })
```
0x4007: Looks like you need to run formatting

zugdev:
> Vercel is private but requested access. Why use vercel and not our continuous deploy?

No difference tbh, I just have it setup for CD in my fork and since the CD here was pending I attached.

static/progressive-web-app.js
zugdev: > Looks like you need to run formatting

This is formatted, the weird indentation is because it's a `.catch` statement if you look closer.

src/home/fetch-github/fetch-issues-full.ts
```diff
@@ -20,9 +20,13 @@ export async function postLoadUpdateIssues() {
     const fetchedIssues = await fetchIssues();
 
     if (issuesAreDifferent(cachedIssues, fetchedIssues)) {
-      await saveIssuesToCache(cachedIssues, fetchedIssues);
+      await saveIssuesToCache(cachedIssues, fetchedIssues); // this handles stale and new issues
       await taskManager.syncTasks();
-      void displayGitHubIssues();
+      if (cachedIssues) {
+        void displayGitHubIssues(undefined, undefined, true); // if there were cached issues skip animation
```
0x4007: I think it makes sense to change the function signature so that you don't have to pass in two undefined? 

Alternatively, if it makes sense, pass in destructured parameters (an object)

0x4007 (changes requested):

0x4007:
If you can fix that I can merge!

0x4007:
When I load the first time there is no animation.

zugdev:
@0x4007 this should be better now, the problem was PWA would serve from cache always and we would need to manually update cache's name to ensure users get distributed updates. I've designed it now so that whenever you open the page you load the newest version into cache on the background, so next time users open the page/app they'll have the newest version.

zugdev:
Still keeping cache first approach since it's much better in performance.

static/progressive-web-app.js
```diff
@@ -55,17 +60,29 @@ self.addEventListener("fetch", (event) => {
 
   event.respondWith(
     caches.match(event.request).then((cachedResponse) => {
-      if (cachedResponse) {
-        return cachedResponse;
-      }
-      return fetch(event.request)
+      const fetchPromise = fetch(event.request)
         .then((networkResponse) => {
```
0x4007, line 64: Can you do async await syntax?

0x4007 (approved):

PR #137, opened 2024-10-23
feat: improve general UI, responsiveness and handle searches in URL

### LINKED ISSUE #122: Add text filter query to URL and filter on render
We should add a parameter to the URL according to the current "Text Filter" component's state. This will allow for people to share specific issues, organizations and repos. An example of how this could look is: `https://devpool.directory/?q=ubiquity-os-marketplace` where on page load a checker function should be called to set the text filter to be `ubiquity-os-marketplace` from the URL. The filter's state should be made persistent which it currently is not, this means that toggling between directory and proposals should keep the filter on.

zugdev:
Resolves #122

### DIFF
--- a/src/home/sorting/sorting-manager.ts
+++ b/src/home/sorting/sorting-manager.ts
@@ -56,9 +56,7 @@ export class SortingManager {
           const gitHubIssue = taskManager.getGitHubIssueById(parseInt(issueId));
           if (!gitHubIssue) return;
           const searchableProperties = ["title", "body", "number", "html_url"] as const;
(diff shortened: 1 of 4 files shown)


zugdev:
@0x4007 this looks sick :)

0x4007:
Perhaps you can make the text filter on top on mobile?

zugdev:
This will require a little extra work, but I can do that. If you wanna merge this one for now, I will refac it later.

0x4007 (approved):

PR #138, opened 2024-10-24
Mobile fix

### LINKED ISSUE #135: Mobile Touch Event List Transparency
image

When scrolling on mobile, after hitting one of the list items with my thumb, it makes that single list item "active" and fades out the rest, even after lifting up thumb.

This behavior is intended for mouse activity. The UX works well on desktop only, but not on touchscreens. 

Disable this behavior for mobile, and test for scrolling in both up and down directions.

obeys:
Resolves #135

### DIFF
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -113,7 +113,7 @@
     transition: 125ms opacity ease-out 62.5ms;
   }
   #issues-container:hover .issue-element-inner {
-    opacity: 1;
+    opacity: 0.5;
   }
   #issues-container.keyboard-selection:hover .issue-element-inner {
     opacity: 1;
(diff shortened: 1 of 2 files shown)


obeys:
erm I think I may have broken something

obeys:
fixed the site breaking css changes!

0x4007 (changes requested):
This is done wrong. The behavior needs to be the same with mouse and disabled on mobile touch only

obeys:
> This is done wrong. The behavior needs to be the same with mouse and disabled on mobile touch only

Could you elaborate more about that? I believe the current implementation handles the mouse hover correctly and disables the fading effect on mobile devices. video below to showcase

https://github.com/user-attachments/assets/c6061831-eaa8-4439-81ad-4f4870626039

0x4007:
I was expecting event handlers and such let me test on computer later but your video looks promising thanks!

0x4007 (approved):
My mistake

PR #140, opened 2024-10-25
feat: handle orgs sublinks (i.e: work.ubq.fi/ubiquity)

zugdev:
Resolves #128

Wait for #136 to be merged before merging this one since it uses PWA.

PR #143, opened 2024-10-30
feat: use org name in header

### LINKED ISSUE #141: White Labeling Header
Extending #128

We can replace the header logo with the GitHub organization avatar and then use the organization name. 

For example

(AAVE logo) AAVE DIRECTORY

Will need to be cautious to handle long organization names like UbiquityOS Marketplace for short width screens!

zugdev:
Resolves #141

This pulls from #140 so merge it first.

Check the video:

https://github.com/user-attachments/assets/b1eeb8b1-fd69-4d9e-842e-5e3de9c5165f

Also am not sure if we should keep DEVPOOL always? Lmk


.gitignore
```diff
@@ -8,6 +8,7 @@ node_modules
 .pnp.loader.mjs
 static/dist
 .env
+.wrangler/
```
0x4007, line 11: You sure about this?

src/home/rendering/render-org-header.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Convert Blob to a URL
+    const logoUrl = URL.createObjectURL(logoBlob);
+
+    const img = document.createElement("img");
+    img.src = logoUrl;
+    img.alt = `${orgName} Logo`;
+    img.id = "logo";
+    img.style.width = "24.469px"; // same size as default SVG
+    img.style.height = "28px";
+    img.style.objectFit = "contain";
```
0x4007: I think it's better to change the css instead of setting directly in typescript.

static/progressive-web-app.js
```diff
@@ -14,75 +14,87 @@ const urlsToCache = [
 // Install event (caches all necessary files)
 self.addEventListener("install", (event) => {
```
0x4007: Async this instead

0x4007:
> Resolves #141
> 
> 
> 
> This pulls from #140 so merge it first.
> 
> 
> 
> Check the video:
> 
> 
> 
> https://github.com/user-attachments/assets/b1eeb8b1-fd69-4d9e-842e-5e3de9c5165f
> 
> 
> 
> Also am not sure if we should keep DEVPOOL always? Lmk

I'm not sure either so let's leave it for now and change later if we need to. 

We have the domain name devpool.directory/ubiquity etc so it makes sense to still reference it. 

It would be interesting to 

1. Have orgs `cname` the domain to "host" on their own domains 
2. Dynamically generate the color theme based on their org avatar but seems difficult hard to do well, and isn't a priority so I think we can't fund that research.

.gitignore
zugdev: Yes, checkout https://github.com/ubiquity/pay.ubq.fi/blob/development/.gitignore

static/progressive-web-app.js
zugdev: like this? `self.addEventListener("install", async (event) => {` not sure why though

zugdev:
> It would be interesting to
> 
> 1. Have orgs `cname` the domain to "host" on their own domains

I have that in mind, yes.

> 2. Dynamically generate the color theme based on their org avatar but seems difficult hard to do well, and isn't a priority so I think we can't fund that research.

This is very cool and is not that hard tbh. We could create an issue with low priority and re-label it when the right time comes.

.gitignore
0x4007: Just because it's there doesn't mean it's correct

zugdev: I wasn't inspired by it: the files in .wrangler are either temporary or state related. Temp files are generated at run time, we also don't want to share kv state so it makes sense to ignore it.

image

static/progressive-web-app.js
```diff
@@ -12,77 +12,89 @@ const urlsToCache = [
...
-    })
+        return cachedResponse || (await fetchPromise);
+      } catch (error) {
+        console.error("[Service Worker] Error handling fetch:", error);
+        return new Response("An error occurred", {
+          status: 500,
+          statusText: "Internal Server Error",
+        });
+      }
+    })()
```
0x4007, line 105: It's strange that you do self invoking functions to do async. Just wrap the entire callback inside an async function like we discussed here and then you can await them like normal. 

Also change catch to try catch. And don't use thenable, use async await.

0x4007 (changes requested):

static/progressive-web-app.js
zugdev: The self invoking functions are needed because event's `waitUntil()` method receives a promise as a parameter - not a function that returns a promise, ([see reference](https://www.w3.org/TR/service-workers/#wait-until-method)). I can only imagine that your idea was to remore the use of `waitUntil` entirely, for instance:

```
// Install event (caches all necessary files)
self.addEventListener("install", async (event) => {
  try {
    const cache = await caches.open(cacheName);
    await cache.addAll(urlsToCache);
    console.log("[Service Worker] Files cached successfully");
  } catch (error) {
    console.error("[Service Worker] Cache failed:", error);
  }
  self.skipWaiting(); // Activate the new worker immediately
});
```

instead of:

```
// Install event (caches all necessary files)
self.addEventListener("install", async (event) => {
  event.waitUntil(
    (async () => {
      try {
        const cache = await caches.open(cacheName);
        await cache.addAll(urlsToCache);
        console.log("[Service Worker] Files cached successfully");
      } catch (error) {
        console.error("[Service Worker] Cache failed:", error);
      }
      self.skipWaiting(); // Activate the new worker immediately
    })()
  );
});
```

The difference between these approaches is that with `waitUntil()` adding assets to cache becomes an asynchronous task, which would allow the task to be completed even after returning inside the event handler. During install this is not a problem, however in a fetch event I want the service worker to return whatever is in cache, but fire an asynchronous task to update the cache with a network response. If we remove `waitUntil()` and just await the network response, we would have to wait for network response to return whatever is in cache, seriously compromising speed in very slow networks. 

Here is a QA using `waitUntil()` to return from cache and asynchronously update cache with network response, in a simulated 10kbps network speed:

[10kbps-network-waitUntil.webm](https://github.com/user-attachments/assets/7429f941-b396-4134-bd0c-4b77726ac743)

Here is a QA awaiting the network response, in a simulated 10kbps network speed:

[10kbps-network-await.webm](https://github.com/user-attachments/assets/344132a9-8d2a-4054-a1dd-58943063792a)

Here are some additional references to clarify:

[What does event.waitUntil do in service worker and why is it needed?](https://stackoverflow.com/questions/37902441/what-does-event-waituntil-do-in-service-worker-and-why-is-it-needed)

[JavaScript how to use event.waitUntil() to wait for an async function call](https://stackoverflow.com/questions/65530092/javascript-how-to-use-event-waituntil-to-wait-for-an-async-function-call)

[Should I use event.waitUntil() or await fetch() or just fetch()?](https://community.cloudflare.com/t/should-i-use-event-waituntil-or-await-fetch-or-just-fetch/56183)

0x4007: Okay seems like you did your research so let's do what you think is right.

0x4007 (approved):

static/progressive-web-app.js
zugdev: That was good feedback, it's very important that we understand each others code practices, as well as document weird patterns. Thanks.

zugdev:
Can we merge #140 and then this one (#143)?

0x4007:
> Can we merge #140 and then this one (#143)?

I don't think order should matter git is supposed to handle this

PR #145, opened 2024-10-30
feat: add some bottom padding to issues-container

### LINKED ISSUE #144: add some more bottom padding to issues-container
The mobile UI is overlapping a little with the last issue, I'll just add some padding 

image

zugdev:
Resolves #144

QA:
image

### DIFF
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -100,6 +100,7 @@
     scrollbar-width: none;
     -ms-overflow-style: none;
     padding: 56px 0;
+    padding-bottom: 90px;
     transition: 0.25s all cubic-bezier(0, 1, 1, 1);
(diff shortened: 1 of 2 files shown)


0x4007:
This doesn't fix. Consider the iOS specific viewport resizing due to the safari toolbar

zugdev:
You mean preview is being cut by footer, so we have 2 bugs, because the issue list was previously hidden under bottom toolbar too. Will add a fix for the preview.

This is what currently is solved:
image

zugdev:
This is very weird, when emulating mobile on desktop it works fine as in screenshot:

image

But it doesn't work in my mobile too. Not only Safari, Chrome and Firefox too. I can take a closer look later, but could create an issue on it too.

PR #147, opened 2024-10-30
feat: add labels to preview when necessary

### LINKED ISSUE #142: Add labels at issue preview body
When we click on an issue we should show the labels - price, time, priority - inside the body specially at mobile view, but I guess it's benefic in Desktop too. This way a user can navigate using the buttons and still get a feel for the issue's characteristics.

Forcefully adding the div as an example:

image

zugdev:
Resolves #142

https://github.com/user-attachments/assets/a6589d66-8e3a-401a-8247-06ec8aef07da


0x4007:
Mixed feelings on how it looks on mobile but the bottom can't be scrolled to all the way at least in iOS now with the labels offsetting the inner height.

0x4007:
I think the labels would look better on the bottom on mobile for sure.

zugdev:
> but the bottom can't be scrolled to all the way at least in iOS now with the labels offsetting the inner height. 

Do you mean you cant scroll to the bottom of the issue list? I've added another PR to add some padding

zugdev:
> I think the labels would look better on the bottom on mobile for sure. 

Will try it!

0x4007:
> > but the bottom can't be scrolled to all the way at least in iOS now with the labels offsetting the inner height. 
> 
> Do you mean you cant scroll to the bottom of the issue list? I've added another PR to add some padding
> 

It should be in this pull because this pull introduced the problem.

0x4007:
https://github.com/user-attachments/assets/e9876e47-544f-4f01-b133-e4663661031f

0x4007:
I just realized that the problem is in production. I think another recent pull introduced the problem.  

The footer bar is too tall now. I must have overlooked it in another pull.

zugdev:
> I just realized that the problem is in production. I think another recent pull introduced the problem.
> 
> The footer bar is too tall now. I must have overlooked it in another pull.

Yeah me too, it was introduced in #137 when toolbar was refactored. #145 is a quick solution.

zugdev:
> I think the labels would look better on the bottom on mobile for sure.

Do you have stacked or side by side in mind? I'll attach some quick illustrations for comparison:

Stacked:

image

image

Side by side:

image

image

Side by side can be good if I can reduce buttons a little, but labels will wrap when too big. 

IPhone XR demo:

image

0x4007:
Side by side but if I had to choose stacked I would make the control buttons on top.

zugdev:
@0x4007 lmk what you think about this.

0x4007 (approved):
Super clean on mobile. Not sure on desktop but I'm assuming you also did well there. 

A bit concerned about those void statements. I'm not sure if an error is thrown if it will be caught.

PR #149, opened 2024-10-31
Client Side Searching

### LINKED ISSUE #119: Client Side Embeddings Search
image

Perhaps we can improve our search experience by:
1. Loading in all the vector embeddings of every issue and associated comments from our database
2. Run the similarity search (i.e. cosine) to rank sort the most relevant issue

If performance is bad running all of these calculations, we can compile to wasm potentially.

shiv810:
Resolves #119 

- Adds a new search system, using a fuzzy matching algorithm combined with Normalized Discounted 
Cumulative Gain (NDCG) for result ranking.


0x4007:
What's some good search terms to test?

I know before was exact string matches. 

I think it would be really useful to segment the search results like:

At the top would be exact matches

And then below a line would say similar results:

shiv810:
> What's some good search terms to test?
> 

A good example for testing would be "Large language models", original implementation would not give results for that. 

> I know before was exact string matches. 
> 
> I think it would be really useful to segment the search results like:
> 
> At the top would be exact matches
> 
> And then below a line would say similar results:
> 
> 

It's, weighted scoring so the exact match should have the highest score and be at the top most position, or should there be two sections one for exact matches and another for similar results.

0x4007:
I think I would feel more comfortable if we can enable fuzzy search so we can compare results over time and tweak the parameters. Perhaps you can add a small checkbox next to the search bar?

Alternatively we could prefix the search with a symbol like ?

?large language model 

And then it will fuzzy search

zugdev:
In current prod search implementation. If you search for "UI", nothing changes because of UI in ubUIquity:

image

To me it's obvious the title should have a bigger weight, we should not consider body URLs and there should be sorting by relevance. You apparently are weighting title, body, etc... I think what could improve your solution greatly is removing URLs from issue's body when scoring. Strange stuff happens because of URL:

image

The UI bug is actually because of matching with URLs.

image

Scoring should eventually account for beginning of words too. But this seems like a good PR, just handling URLs should fix a lot of weird results. If you find correct weights and pre-processing, the scoring algo will be good enough so exact results will always be on top.

zugdev:
TLDR: I like the weights idea and I like the relevance sorting. We need to clean issue's body to remove URLs, potentially other stuff too, such as code blocks. we could consider beginning of words as a stronger score in an exponential curve, so that if we have issues "banana" and "analphabet", searching "ban" leads to "banana" and "ana" leads to "analphabet".

zugdev:
This looks great, please try removing the URLs before doing the exponential substring scoring, perhaps that'd be overkill.

shiv810:
Fuzzy search only works with "?" prefixed searches; otherwise, heuristic search is used. URLs are not used for search content anymore. There's an exponential boost for word beginnings, so words matched with similar starting letters would score higher.

Some examples:
- "ban" → Strongly matches "banner" and "banana" (word starts)
- "ana" → Strongly matches "analysis" (word start), weakly matches "banana" (middle match)
- "?ban" → Enables fuzzy search for "ban"-like words
- "database schema" → Matches text while ignoring URLs in the content

zugdev:
URL matching seems to work now!

image

Take a look at [Collaborator Gating Based On Label](https://83baca04.devpool-directory-ui.pages.dev/?search=ui&issue=2559612103), as can be seen in the image it's ranked second, though it has nothing to do with UI - probably because it has "Ubiquity" written two times inside a code block. I see you wrote code to ignore code blocks, might not be properly ignoring. I believe **Update UBQ Farming UI** should be placed higher too, perhaps increase title's weight even more. Though the micro adjusments, this looks very good: nice job.

shiv810:
The results are not sorted by score, as that would conflict with the `sorting-manager`. So, the second entry has a relevance weight of 0.07, while the third has a score of 0.67.

zugdev:
Interesting. Currently in prod, sorting just ignores what's in the text field, so if you search and sort it will show not only the filtered by search ones, but all existent issues anyway. Therefore it wouldn't really conflict, but we should make it clear by clearing the text field when user presses a sorting button. Another option is we let it be and never sort by relevance, allowing sorting to act upon search-filtered issues. @0x4007 RFC

shiv810:
The results are sorted by default by their relevance score. If the you switch to a sorting method, the text box would be cleared. Similarly, if you perform a search while using a sorting method, the sort button will reset.

zugdev:
That's great! Apparently it's very good, I'll review code in depth and circle back.

shiv810:
@0x4007 @zugdev Can you take a look at this pull? I think it's ready for review, but I can't request a review here for some reason.

src/home/sorting/sorting-manager.ts
```diff
@@ -53,19 +54,46 @@ export class SortingManager {
 
     const filterIssues = () => {
       try {
-        const filterText = textBox.value.toLowerCase();
+        const filterText = textBox.value;
```
zugdev: I believe you should not remove this `.toLowerCase()` statement. Because of this removal if you search for:

1. "pool"
image

2. "POOL"
image

you get different results.

zugdev:
That's all for me, the rest looks good.

zugdev:
I have subbed to notifications in this PR, I'll approve once you respond the review.

shiv810:
> I have subbed to notifications in this PR, I'll approve once you respond the review.

I’m not sure if you’ve left a review, but I don’t see any comments on the PR.

zugdev:
> I’m not sure if you’ve left a review, but I don’t see any comments on the PR.

It was pending, pardon.

src/home/sorting/sorting-manager.ts
shiv810: Fixed that.

zugdev (approved):

zugdev:
@0x4007 let me know when you approve this merge.

0x4007:
image

Can't type anything

zugdev:
Neither can I, @sshivaditya2019 this last commit broke it.

shiv810:
That should be fixed. I'm not sure why this is happening, but I was able to replicate the issue only on the Cloudflare builds; the local versions worked fine on the same browser(mobile).

zugdev:
This does not fix it, it just doesn't reset sorting when you search because the if fails. The error is in a `_resetSortButtons` call

image

so investigate where you call this function. QA:

image

zugdev:
ps: this is not the cause

I did some quick looking and it might be because `filterIssues` runs on load, given the observer:

https://github.com/ubiquity/work.ubq.fi/blob/bd41261088d2194c70602a18bdb5323cc83dee7c/src/home/sorting/sorting-manager.ts#L74-L81

And the SortingManager only is created on: 

https://github.com/ubiquity/work.ubq.fi/blob/bd41261088d2194c70602a18bdb5323cc83dee7c/src/home/sorting/generate-sorting-buttons.ts#L6-L12

Even though the function above runs in line 20 of `home.ts` on cloud it might run after filterIssues which implies undefined elements to reset.

zugdev:
@sshivaditya2019 [This might be the commit that broke it](https://github.com/sshivaditya2019/work.ubq.fi/commit/5d905d9ffae7bb0682fcf583409271a67aa7f815)

shiv810:
I am not sure, if this is usual/related to this pull, but I am not able to open the website on Safari (iOS) and Private Tab in Brave browser (iOS).  I am getting `Failed to store record in an IDBObjectStore: BlobURLs are not yet supported.`

shiv810:
> @sshivaditya2019 [This is the commit that broke it](https://github.com/ubiquity/work.ubq.fi/commit/bd41261088d2194c70602a18bdb5323cc83dee7c)

Are you sure it was this commit ? It doesn't seem like it made any changes to the files in this pull request. Also, I haven't pulled in any changes from the development branch

zugdev:
> Are you sure it was this commit ?

[8395fd4](https://04e5dd6a.devpool-directory-ui.pages.dev/) this deployment worked as expected.

[5d905d9](https://bc52aa76.devpool-directory-ui.pages.dev/) this one didn't. Since the commit I mentioned was the one that triggered the deploy I pointed it out. Could be on merge too.

zugdev:
Try removing the toLowerCase and push it please.

zugdev:
Please pull from this repo's development branch and merge in your `devsim` branch.

src/home/sorting/sorting-manager.ts
```diff
@@ -55,17 +59,48 @@ export class SortingManager {
       try {
         const filterText = textBox.value.toLowerCase();
         const issues = Array.from(issuesContainer.children) as HTMLDivElement[];
+
+        // Reset any active sort buttons when searching
+        if (filterText) {
+          if (this) {
+            this._resetSortButtons();
```
zugdev: `this` isn't valid here because of scope, if you make filterIssues an arrow function it will work:

change from `function filterIssues() {` to `const filterIssues = () => {`

shiv810: Isn't `filterIssues` an arrow function? I don't think it's relevant anymore. I've moved it to `textBox.addEventListener`, and by the time it gets there, `SortingManager` should already be defined.

.gitignore
```diff
@@ -11,3 +11,12 @@ static/dist
...
 cypress/videos
+
+# Wrangler
+.wrangler
+
+# vscode
+.vscode
+
+# husky
+.husky
```
zugdev: Don't ignore `.husky`

zugdev (changes requested):

zugdev (approved):

shiv810:
@0x4007 Is there anything still pending for this PR? I can make the necessary changes for that.

zugdev:
> @0x4007 Is there anything still pending for this PR? I can make the necessary changes for that.

He is currently busy at a conference. I have tested this a lot, there is only one case I found a small bug. When you

1. Open https://5f3d89d0.devpool-directory-ui.pages.dev and search for "pool"

2. Copy the URL of this search

3. Open in a new tab, the order is different

I guess that's the last thing. Once you handle that I'll merge it.

shiv810:
It should be fixed in `94533de`. I changed the way the results were sorted.

zugdev:
The order is fixed, but you should skip loading animation if search was derived from the URL as currently in prod:

Open in prod, it's smooth: https://devpool.directory/?search=aa

https://github.com/user-attachments/assets/61dda2bc-6a71-419c-9c1b-feb192d39345

In `94533de` it lags: https://ba848e44.devpool-directory-ui.pages.dev/?search=pool

https://github.com/user-attachments/assets/e27a3ea8-6ba9-4d64-8ad9-5849af65634a

This is because the animation is taking into account the non-displayed in between issues. I will pinpoint where you need to modify in code review in a sec.

src/home/sorting/sorting-manager.ts
```diff
@@ -50,22 +54,36 @@ export class SortingManager {
...
-          if (!gitHubIssue) return;
-          const searchableProperties = ["title", "body", "number", "html_url"] as const;
-          const searchableStrings = searchableProperties.map((prop) => gitHubIssue[prop]?.toString().toLowerCase());
-          const isVisible = searchableStrings.some((str) => str?.includes(filterText));
-          issue.style.display = isVisible ? "block" : "none";
-        });
+        // Get issue IDs and search results
+        const issueIds = issues
+          .map((issue) => issue.children[0].getAttribute("data-issue-id"))
+          .filter((id): id is string => id !== null)
```
zugdev: why this filter and second map?

src/home/sorting/sorting-manager.ts
```diff
@@ -50,22 +54,36 @@ export class SortingManager {
...
+          .map((id) => parseInt(id));
+
+        const searchResults = taskManager.issueSearcher.search(filterText, issueIds);
+        // If there's a search term, sort by relevance
+        if (filterText) {
+          issues
+            .sort((a, b) => {
```
zugdev: would you mind writing a function in another file to abstract this code section? this is how we do it for other sorting criteria:

image

src/home/sorting/sorting-manager.ts
```diff
@@ -50,22 +54,36 @@ export class SortingManager {
...
+              const scoreA = parseFloat(a.getAttribute("data-relevance-score") || "0");
+              const scoreB = parseFloat(b.getAttribute("data-relevance-score") || "0");
+              return scoreB - scoreA; // Sort in descending order of relevance score
+            })
+            .forEach((issue) => {
+              const issueId = issue.children[0].getAttribute("data-issue-id");
+              if (!issueId) return;
```
zugdev: Just add `issue.classList.add("active");` above `if (!issueId) return;` to fix that loading animation problem

src/home/sorting/sorting-manager.ts
shiv810: The IssueSearch already contains all of the issues; this only returns the issue IDs in their current order. It could be extended to include features for sorting and searching, essentially allowing us to search and sort (like Price, Time)

src/home/sorting/sorting-manager.ts
shiv810: Moved the`filterIssues` to a separate file.

zugdev:
Hey, I know this has been a long process, sorry. But latest changes introduced a "unsorting" problem:

1. Initial order is not restored when unsorted:

Initial state:
image

State after clicking to sort by time until unsorting:
image

2. If you sort and then search, even though the sort button is visually reset, the sort is still applied

[Screencast from 14-11-2024 17:37:10.webm](https://github.com/user-attachments/assets/9671ae08-dc01-4b65-9213-cf3f5f737644)

src/home/sorting/filter-issues.ts
```diff
@@ -0,0 +1,33 @@
+import { taskManager } from "../home";
+
+export function filterIssues(textBox: HTMLInputElement, issuesContainer: HTMLDivElement) {
```
zugdev: please rename this file and functions to: filter-issues-by-search.ts and filterIssuesBySearch respectively.

shiv810:
> Hey, I know this has been a long process, sorry. But latest changes introduced a "unsorting" problem:
> 

This has been fixed in `aae1a02`. I have rewritten the `filterIssues` to make it more consistent with how issues are usually rendered for other sorting methods.

zugdev:
By QA this looks good, I'll simplify some code in another PR. Thank you, for your responsiveness, it was great working with you.

PR #150, opened 2024-11-01
feat: fix flickering on login

zugdev:
This event handler was reloading because callback uses popstate, it's not even a good feature adding issues to URL history so let's just rollback this feat. To be clear, what this removes is that previously as you navigated through issues we would push them to history so when user popped state (clicked to go back a page) it would load the previous issue.


zugdev:
One little bugged flow remains: login, logout, login, without ever refreshing.

zugdev:
@0x4007 ready to review

0x4007:
@ubiquityos Can you link the issue/pull we are reverting for additional context?

PR #157, opened 2024-11-05
fix: code blocks in preview

### LINKED ISSUE #155: expand code blocks in preview
Code blocks in issue previews are currently barely readable, take the image from [this preview](https://devpool.directory/?issue=2208105478) as an example:

image

On GitHub it looks like this:

image

Which is much more readable.

dawidsawa:
Resolves #155

### DIFF
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -519,6 +519,7 @@
     position: relative;
     display: inline-block;
     width: 100%;
+    box-sizing: border-box;
     overflow-x: auto;
     padding: 16px;
     background-color: #7f7f7f20;
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -519,11 +519,8 @@
     position: relative;
     display: inline-block;
     width: 100%;
-    overflow: scroll;
-    height: 48px;
   }
   .preview pre code {
-    position: absolute;
     font-family: monospace;
   }
   .preview ol,
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -519,7 +519,7 @@
     position: relative;
     display: inline-block;
     width: 100%;
-    overflow-x: scroll;
+    overflow-x: auto;
     padding: 16px;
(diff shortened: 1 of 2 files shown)


dawidsawa:
Before
before

After
after

zugdev:
Please add `overflow-x: scroll` in code block, to handle issues like [this one](https://aafe466c.devpool-directory-ui.pages.dev/?search=Improve+Cloudflare+Worker&issue=2623315704)

image

Also consider adding a slightly darker background color for the code block.

dawidsawa:
Given the background is already very dark adding even an absolute black wouldn't be much visible anyway, so perhaps using a lighter background [like here](https://37843ffb.devpool-directory-ui.pages.dev/?search=Improve+Cloudflare+Worker&issue=2623315704):
scroll-bg

zugdev:
Please consider the inverted style too because when browser is in light mode the following happens:

image

Also:

1. Add some padding so text is not too close to the code block's border. Notice that the biggest difference from our code block to GitHub's is padding:

image

image


2. Make the font size inside the code block 12px, which is a tiny bit smaller, allowing for more content to be shown.

dawidsawa:
How about now:

dark

light

zugdev:
That's better, one last thing check [this issue](https://de788862.devpool-directory-ui.pages.dev/?issue=2595388408). The code block is unnecessarily overflowing to the right:

image

We want it to look like:

image

It seems it's because of the added padding. If you know a good solution that doesn't involve calc -32px, please use it.

dawidsawa:
EDIT: my bad, that did not fix it.

https://a642e753.devpool-directory-ui.pages.dev/?issue=2595388408

pre-overflow-fixed

zugdev:
Looks good! Not sure about hard-coded colors tho, @0x4007 RFC

zugdev (approved):

zugdev:
This is how it looks now:

image

We could possibly look into a coloring lib in the future.

PR #163, opened 2024-11-08
feat: add kv binding in actions for referral tracker

### LINKED ISSUE #151: Add variables into wrangler deployment script
All of the KV setup is automated via the Action deployment script, these variables should be described there. [This](https://github.com/ubiquity/work.ubq.fi/pull/123/files#diff-b1b9719b6eae4103d520f1e902420f5073b5e18a5a47aa73feed71a037557c98R9) should not have been commited as such. Example from the plugin: https://github.com/ubiquity-os/plugin-template/blob/development/.github/workflows/worker-deploy.yml#L34

By hardcoding the bindings if we change org / try to deploy on our own, it will break.

_Originally posted by @gentlementlegen in https://github.com/ubiquity/work.ubq.fi/issues/123#issuecomment-2440249980_

zugdev:
Resolves #151

@gentlementlegen I know this ain't right as of now, but I'd love some feedback. I've based myself of https://github.com/ubiquity-os/ubiquity-os-kernel/.


gentlementlegen:
@zugdev that's the spirit, but removing `ubiquity/cloudflare-deploy-action@main` in favor of `cloudflare/wrangler-action@v3` might have side effects. I don't know all the work that happens behind the scenes but I think the first Action takes care of previews and building URLs and such. @0x4007 might know more. You'd want to take a look at the source code : https://github.com/ubiquity/cloudflare-deploy-action/. It is maybe even better to actually update that package directly.

0x4007:
I don't know more.

zugdev:
I'll test it out.

zugdev:
@0x4007 can you pls give me actions permission for me to try running this deploy workflow file?

0x4007:
> @0x4007 can you pls give me actions permission for me to try running this deploy workflow file?

I missed this. Are you asking to be admin? I'll do that now

wrangler.toml
```diff
@@ -4,14 +4,11 @@ compatibility_date = "2024-10-23"
 
 pages_build_output_dir = "./static"
 
-[[kv_namespaces]]
-binding = "KVNamespace"
-id = "0a6aaf0a6edb428189606b116da58ef7"
+[env.dev]
+[[env.dev.kv_namespaces]]
+binding = "kv"
+id = "9cd5fdf01971402390dbe9e95cc078d3"
```
gentlementlegen, line 11: This is what we are trying to resolve, it should not be here, the deployment will fail.

zugdev: I thought this was what we did at kernel

https://github.com/ubiquity-os/ubiquity-os-kernel/blob/development/wrangler.toml

image

zugdev:
Managed to make it work by copying kernel, here is QA:

https://github.com/ubiquity/work.ubq.fi/actions/runs/12944850895

@gentlementlegen can you take a quick look?

wrangler.toml
zugdev: We gotta have the production fields else update toml fails cause it doesnt find a field to change

gentlementlegen: The kernel does not use any of these, they actually should get removed.

Yes but I think the fields can be empty, like for plugins. Example: https://github.com/ubiquity-os-marketplace/command-start-stop/blob/development/wrangler.toml

PR #165, opened 2024-11-13
feat: fix live reload

### LINKED ISSUE #164: fix live reload
We've recently changed this repo's `package.json` start command to use wrangler pages because of the referral tracker kv implementation. 

```json
"start": "wrangler pages dev --port 8080"
```

This killed the live reload feature, and https://github.com/cloudflare/workers-sdk/issues/4824 we have to pass in a directory after wrangler pages dev to hear changes from. However, passing in `./` wont work since it only reads inside the received directory (non-recursive).

zugdev:
Resolves #164


package.json
```diff
@@ -9,8 +9,9 @@
     "node": ">=20.10.0"
   },
   "scripts": {
-    "start": "wrangler pages dev --port 8080",
-    "watch": "tsx build/esbuild-server.ts",
+    "start": "concurrently \"yarn start-wrangler\" \"yarn watch\"",
```
0x4007:
```suggestion
    "start": "run-s start-wrangler watch",
```

We use npm-run-all for exactly situations like these

package.json
```diff
@@ -46,6 +47,7 @@
     "@types/node": "^20.10.0",
     "@typescript-eslint/eslint-plugin": "^6.13.1",
     "@typescript-eslint/parser": "^6.13.1",
+    "concurrently": "^9.1.0",
```
0x4007:
```suggestion
```
Remove and uninstall so lock file is up to date

0x4007 (changes requested):

package.json
zugdev: gotta use run-p actually

0x4007 (approved):

PR #166, opened 2024-11-22
feat: fix style on tables

### LINKED ISSUE #159: properly style tables in preview
A table in an issue description gets correctly converted into an HTML table element, however the style is not as neat as GitHub's. Take [this issue](https://37843ffb.devpool-directory-ui.pages.dev/?issue=2489452377) as an example:

Ours:
image

GitHub's:
image

We should improve this, I believe it's doable with CSS only.

vichydra:
Resolves #159

### DIFF
--- a/static/style/special.css
+++ b/static/style/special.css
@@ -1,8 +1,12 @@
 :root {
   --dark-background: #000210;
   --dark-background-half: #00021080;
+  --alt-dark-background: #808080;
+  --alt-dark-background-half: #80808020;
   --light-background: #f8f8f8;
   --light-background-half: #f8f8f880;
+  --alt-light-background: #7f7f7f;
+  --alt-light-background-half: #7f7f7f20;
 }
 
 @media (prefers-color-scheme: dark) {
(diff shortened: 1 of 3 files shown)


vichydra:
Old dark mode:

image

New dark mode:

image



Old light mode:

image


New light mode:

image

zugdev:
Can you please use the same colors as in the code block for the alt color in the table?

zugdev:
Please add bottom-padding to the table, it's not a footer only thing.

image

static/style/special.css
```diff
@@ -1,8 +1,12 @@
 :root {
   --dark-background: #000210;
   --dark-background-half: #00021080;
+  --alt-dark-background: #808080;
```
zugdev, line 4: Thanks for the hardcode refac!

zugdev:
Looks good on mobile too:

image

I've noticed the text looks blurry, but it seems to be in prod, I'll fix it.

zugdev:
The blurry text is Opera's fault, it's about how they render text when DevTools is open. Since it only affects when DevTools is open it's not a relevant bug and we should not worry.

PR #170, opened 2024-11-29
feat: repos in search

### LINKED ISSUE #169: add a repo searching scorer
When searching for a repo, the expected behavior is to see all issues within it. Currently we don't support that:

image

We should add that and perhaps for orgs too, although you can already filter by org by appending `/org-name` to the URL:

image

zugdev:
Resolves #169

QA:

image

image

### DIFF
--- a/src/home/issues-search.ts
+++ b/src/home/issues-search.ts
@@ -9,6 +9,7 @@ export class IssueSearch {
     body: 0.25,
     fuzzy: 0.25,
     meta: 0.125,
+    repo: 0.1,
   };
 
   private readonly _config: SearchConfig = {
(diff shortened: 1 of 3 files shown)


zugdev:
@0x4007 I've added org scoring to the search too.

Previously:
image

Now:
image

The weight to repo and org searches is the lowest of them all, so search still prioritizes body, title and metadata.

zugdev:
Multi-term search is also enabled:

image

0x4007:
https://404ad59a.devpool-directory-ui.pages.dev/ubiquity-os-marketplace?search=Text-vector-embeddings

image

- Yeah let's do rounded avatar. I suppose it might look best to match the user avatar to the org avatar on the left in terms of width, height, and placement.  
- check if you can vertical align center on the devpool directory title. Looks a bit low compared to the org icon on the top left 
- disable auto capitalization and spell check in the text search box

0x4007 (approved):

PR #181, opened 2024-12-04
feat: fix alignment

### LINKED ISSUE #180: fix alignment
align this:
image

zugdev:
Resolves #180

image
image

### DIFF
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -363,7 +363,7 @@
     width: 100%;
     height: 48px;
     top: 0;
-    padding: 0 5px;
+    padding: 0 8px;
     backdrop-filter: blur(8px);
     display: flex;
     border-bottom: 1px solid #7f7f7f20;
@@ -375,7 +375,7 @@
   }
   #authentication {
     position: absolute;
-    right: 5px;
+    right:8px;
     cursor: pointer;
     vertical-align: middle;
     flex: 0 0 auto;
@@ -795,9 +795,6 @@
   #view-toggle {
     flex: 0 0 auto;
     appearance: none;
-  }
-  label[for="view-toggle"] {
-    margin-left: 8px;
   }
   #bottom-right {
     position: fixed;
--- a/static/style/inverted-style.css
+++ b/static/style/inverted-style.css
@@ -360,18 +360,22 @@
   }
   .toolbar {
     position: fixed;
-    width: 100vw;
+    width: 100%;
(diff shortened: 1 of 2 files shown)


static/style/inverted-style.css
```diff
@@ -360,18 +360,22 @@
...
     -webkit-backdrop-filter: blur(8px);
     align-items: center;
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
+    box-sizing: border-box;
   }
   #authentication {
     position: absolute;
-    right: 0;
+    right: 5px;
```
0x4007: Doesn't seem right should be a multiple of 4px

static/style/style.css
```diff
@@ -360,18 +360,22 @@
   }
   .toolbar {
     position: fixed;
-    width: 100vw;
+    width: 100%;
     height: 48px;
     top: 0;
+    padding: 0 5px;
```
0x4007: Same

static/style/style.css
```diff
@@ -360,18 +360,22 @@
...
     -webkit-backdrop-filter: blur(8px);
     align-items: center;
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
+    box-sizing: border-box;
   }
   #authentication {
     position: absolute;
-    right: 0;
+    right: 5px;
```
0x4007: Same

0x4007:
Can you try to match the spacing of the left side more closely

static/style/inverted-style.css
zugdev: there is a 1px border in the issue display

0x4007:
The fastest way to handle this is if I just do it but basically the right side spacing does not match the left side (devpool directory header) spacing.

zugdev:
I'll do it

zugdev:
should be good now @0x4007

zugdev:
when I get some time later I will remove absolute from authenticated and the padding will work for both view-toggle and authenticated. I also plan in handling the username in a better way

zugdev:
image

0x4007 (approved):
Looks better when I tested from mobile. Will test on desktop another day.

0x4007:
> when I get some time later I will remove absolute from authenticated 

I did this as a fix because when it would load it would move elements on the header bar but if we can set a fixed width (the usernames are variable length) then it should render fine.

PR #182, opened 2024-12-04
feat: use daemon-pricing

zugdev:
use daemon pricing instead of assistive pricing worker

### DIFF
--- a/.github/.ubiquity-os.config.yml
+++ b/.github/.ubiquity-os.config.yml
@@ -6,23 +6,35 @@ plugins:
         with:
           allowPublicQuery: true
   - uses:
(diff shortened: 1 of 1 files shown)


0x4007 (changes requested):
Collaborators only should match our org settings

zugdev:
I believe it does match
https://github.com/ubiquity/.ubiquity-os/blob/f94809ffc2b3934ff8b8a0035ed47eea56e43f54/.github/.ubiquity-os.config.yml#L32-L63

0x4007:
Oh I confused line 18 required labels to start!

0x4007 (approved):

PR #189, opened 2024-12-23
feat: add availability filter for the issues

### LINKED ISSUE #185: Show Available and In Progress Tasks
It seems that there are two popular use cases of this UI:

1. Finding new available tasks to work on
2. Finding any in progress tasks

I think that by default we should show only the available tasks and we should have a toggle, like a filter button, that shows all of the tasks

_Originally posted by @0x4007 in https://github.com/ubiquity/work.ubq.fi/issues/148#issuecomment-2556836744_

---

This also implies that we need to fix the label logic on the backend. 

I think that descriptive labels can be helpful for further filtering on the UI. 

Some ideas for states to consider:

- available
- assigned
- gating based on priority level or time 

However gating is more complex and should be handled as part of a separate task because it requires interpreting the command-start-stop configs, and the user authentication to determine whether they can start it or not.

koya0:
Resolves #185 

# QA

https://github.com/user-attachments/assets/1ad5283a-719a-42d7-89cb-b6fd6c969e69

### DIFF
--- a/src/home/fetch-github/fetch-and-display-previews.ts
+++ b/src/home/fetch-github/fetch-and-display-previews.ts
@@ -13,6 +13,12 @@ export type Options = {
   ordering: "normal" | "reverse";
 };
 
+export let isFilteringAvailableIssues = true;
+
+export function swapAvailabilityFilter() {
+  isFilteringAvailableIssues = !isFilteringAvailableIssues;
+}
+
 // start at view based on URL
 export let isProposalOnlyViewer = new URLSearchParams(window.location.search).get("proposal") === "true";
 
@@ -77,12 +83,10 @@ function filterIssuesByOrganization(issues: GitHubIssue[]): GitHubIssue[] {
 export async function displayGitHubIssues({
   sorting,
   options = { ordering: "normal" },
-  filterAvailableIssues = true,
   skipAnimation = false,
 }: {
   sorting?: Sorting;
   options?: { ordering: string };
-  filterAvailableIssues?: boolean;
   skipAnimation?: boolean;
 } = {}) {
   await checkCacheIntegrityAndSyncTasks();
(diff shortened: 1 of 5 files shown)


zugdev:
1. You need to change in `style.css` the portrait or max-width to account for the new space. To avoid the following:

Screenshot 2024-12-24 at 18 26 55

2. You should give a little more padding to the new button, notice the difference:

Screenshot 2024-12-24 at 18 28 13

3. It's preferable that you allocate the same width for both states of the button instead of changeable width, as the later generally provokes responsiveness issues. Also a general better text could be "Unassigned"/"All Issues", i.e: something shorter.

4. The mobile view is very broken. I think you could split the space between search bar and it in a 3/1 flex split, if that works. Stacking vertically either makes it very convoluted, confusing and also violates the bottom-bar max height (that could be changed, but I dislike the approach)

Screenshot 2024-12-24 at 18 31 14

I'd create a div that wraps search box and the new button has `display: flex`, then in search box I'd do `flex: 3`, and in button/label `flex: 1`. Be careful of how it impacts the desktop view.

zugdev:
LGTM, the only thing I'd change is I'd show all issues when we are displaying all issues and unassigned when displaying unassigned as we do with directory/proposals.

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -76,31 +77,37 @@ function filterIssuesByOrganization(issues: GitHubIssue[]): GitHubIssue[] {
 export async function displayGitHubIssues({
   sorting,
   options = { ordering: "normal" },
+  filterAvailableIssues = true,
   skipAnimation = false,
 }: {
   sorting?: Sorting;
   options?: { ordering: string };
+  filterAvailableIssues?: boolean;
```
zugdev: akshually I disagree with this being passed as an argument cause if u forget to properly pass it will do wrong i.e: display on search. it'd be best to have a global as we do with proposals!

zugdev:
i'd love if u could solve the little window between 1050px and portrait where bottom bar doesnt appear. this happens cause we have two display modes:

1. portrait OR max_width 1050px

2. landscape AND width > 800px

bonus problem, but I think just deleting 2 solves :p

zugdev:
thanks, @0x4007 up 2 u

koya0:
Waiting review from @0x4007

koya0:
> @koya0, this task has been idle for a while. Please provide an update.
> 
> 
> 

Waiting @0x4007 for review

0x4007:
image

The left and right padding should match on both rows

I also wish it was more intuitive to understand if it's currently showing everything or just the assigned issues  

But unfortunately I think it would be out of scope to add this: we can show some type of status icon on each task row that indicates if it's available or not. It would be immediately clear what the current filter view state is with this.

0x4007 (approved):
I trust that you can fix the padding situation and then merge this in after that's addressed. Thanks.

koya0:
@0x4007 tks for the feedback

0x4007:
image

Still shifted over https://7e57eb65.devpool-directory-ui.pages.dev

koya0:
> image
> 
> Still shifted over https://7e57eb65.devpool-directory-ui.pages.dev

Is shifted only in light mode
image
image
I dont know why, but apparently the div needs more margin-left in light mode, i will fix it.

koya0:
my localhost is aligned but https://6eb4b888.devpool-directory-ui.pages.dev/ is not

localhost:
image

koya0:
i guess i cant make changes only for light mode, I reopened the terminal and ran yarn start and an auto change was made in the code to set margin-left equal to the style.css file

zugdev:
ya thats weird, let me take a look

zugdev:
add an and color scheme condition to orientation queries in css

koya0:
i guess now its fine @zugdev @0x4007

zugdev:
good

Issue #109, opened 2024-10-07
ci: fix cypress e2e tests


rndquu:
Check [this](https://github.com/ubiquity/work.ubq.fi/actions/runs/11216977982/job/31177473349) failing cypress e2e testing workflow.

As far as I understand it fails because now `work.ubq.fi` [fetches](https://github.com/ubiquity/work.ubq.fi/blob/23d04fb764b63e8445453e2e9ff44b84afbce624/src/home/fetch-github/fetch-issues-full.ts#L5) github issues from the [devpool-issues.json](https://github.com/ubiquity/devpool-directory/blob/79330a6d55e22543a3bb102817e5075cb4dc4d5b/devpool-issues.json) while e2e tests [mock](https://github.com/ubiquity/work.ubq.fi/blob/23d04fb764b63e8445453e2e9ff44b84afbce624/cypress/e2e/devpool.cy.ts#L18-L23) regular github API calls.

What should be done:
- refactor [e2e tests](https://github.com/ubiquity/work.ubq.fi/blob/23d04fb764b63e8445453e2e9ff44b84afbce624/cypress/e2e/devpool.cy.ts) according to the current logic

rndquu:
@gentlementlegen pls check if this issue is relevant

zugdev:
I think it makes sense based on previous issues I've tackled.

0x4007:
> > [!CAUTION]
> > Error: Error: 401 Incorrect API key provided: sk-or-v1*************************************************************bfbc. You can find your API key at https://platform.openai.com/account/api-keys.
> 
> 
> 

@Shiv810 we got a straggler here could you do the same as yesterday please.

0x4007:
> Tip
> 
> Results are being retried

@gentlementlegen looks broken when we switch to OpenRouter

gentlementlegen:
Seems the error is the following: https://github.com/ubiquity-os-marketplace/text-conversation-rewards/actions/runs/13847832711/job/38749711152#step:3:1992

I am not sure about what OpenRouter returned, seems like an empty result.

0x4007:
We need to increase OpenRouter reliability. The good news is that we are almost fully migrated over to OpenRouter with all of our plugins.

ariesgun:
@0x4007, can the rewards be generated now?

PR #190, opened 2024-12-27
fix: adds scraper logic and token handling

### LINKED ISSUE #187: Scraper: Post Login with Github
- [x]  Get User Access Token (Update OAuth permissions ?)
- [x]  Add the new issues for the user using the token

```
We need to populate our matchmaking database. We "register" users in our system when they "Sign in with GitHub" on https://devpool.directory/

```

_Originally posted by @0x4007  in [#56](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/issues/56#issue-2717054509)_

shiv810:
Resolves #187 

- Adds the scraper logic
- Fetches if the data in the DB is older than 24 hrs


0x4007 (approved):
Seems generally fine changes not sure how can be QA'd

Was wondering if another approach makes sense such as only copying with high precision based on when an issue is closed as complete and only that is copied over for example. But if we are hunting down historical issues then obviously this isn't a solution

My concern is redundantly finding the same historical issues and doing the same redundant search and scrape work. I am wondering if there is a better approach to this problem.

shiv810:
@zugdev There are two locksfiles `bun.lockdb` and `yarn.lock`, is that intended, which one is suppoed to be used ?

shiv810:
> Seems generally fine changes not sure how can be QA'd

I'll create a demo video to walk through the flows.

Right now, we store the timestamp of the last scrape for the user, I think it should be possible to use this timestamp as a filter to check for any newly closed issues marked as completed. If there are any, we can proceed with the scrape; otherwise, we can skip the process for that user.

zugdev:
> @zugdev There are two locksfiles `bun.lockdb` and `yarn.lock`, is that intended, which one is suppoed to be used ?

The bun.lockb one is related to Cloudflare Wrangler, the main one is yarn.lock. Use the later.

0x4007:
Perhaps we should consolidate all to bun? @gentlementlegen rfc

gentlementlegen:
Ideally we should use bun everywhere for consistency, except if there is a reason not to. It matters mostly because when testing packages locally, `bun link` and `yarn link` are not interchangeable (at least from my experience).

shiv810:
Is there a way to add secrets in the final build without exposing them in `dist`? I’m thinking of creating a pg function to handle this or using Cloudflare KV.

gentlementlegen:
Yes, you can store and pull them from the GitHub Action environment during deployment.

shiv810:
> Yes, you can store and retrieve them from the GitHub Actions environment during deployment.

You’ll still be able to view the values in the `home.js` file after the final build. Is that fine? I think it’s fine for the `SUPABASE URL` and `ANON KEY`, but storing the `VOYAGE API KEY` in plain text doesn’t seem secure.

gentlementlegen:
Why would it be in the final build? It should not be there.

shiv810:
The scraper logic is executed when a user logs in on the client side. So, we will need voyage API key for embeddings.

gentlementlegen:
Ha yes I see now, I was missing the context sorry. We should definitely not expose that API key on the client side. Sadly, we don't use SSR so I don't know how we can handle this except having some API pour that we host ourselves somewhere, or storing it in cookies?

0x4007:
Make a worker endpoint. We have a setup like this for pay.ubq.fi that's related to the cards. Check its code. 

@EresDev rfc

EresDev:
> Make a worker endpoint. We have a setup like this for pay.ubq.fi that's related to the cards. Check its code.
> 
> @EresDev rfc

Yes, it looks like VOYAGEAI part needs a backend. You can [get started](https://developers.cloudflare.com/pages/functions/) with pages functions.  

We store API keys in cloudflare worker secrets. However, one similar thing that ubiquity-os-kernel [does differently](https://github.com/ubiquity-os/ubiquity-os-kernel/blob/bb72c0adf9da421cbe73a2a53efa31eb71140e2d/.github/workflows/worker-deploy.yml#L43) is store the API keys in Github, but pushing it to cloudflare during worker deployment. This helps with managing the secrets in one place (GitHub) and we probably will do the same soon in `pay.ubq.fi` if possible.

shiv810:
> We store API keys in cloudflare worker secrets. 

We can store API keys in Cloudflare KV, but I don't think this is the best approach, as we could quickly hit the read limit. A better option might be [Supabase Functions](https://supabase.com/docs/guides/functions/quickstart), which would allow us to read directly from the Supabase Vault.

gentlementlegen:
You can have environment secrets in cloud flare that do not count against any quota. We can upload then from Github environment secrets, like we do for most plug-ins.

shiv810:
QA:

Screenshot 2025-01-04 at 11 51 53 AM


Mean Stats:

- CPU Wall time: 3.2ms
- Sub Requests: 16 (For About 61 Issues)

@zugdev could you add the supabase service role key to the secrets under `SUPABASE_KEY` ? That's need for adding the scraped results back to the table. Please ping me in telegram for the value.

shiv810:
@0x4007 

QA:

https://scraperui.work-ubq-fi-50d.pages.dev/


You should be able to see the scraper function request, in the Network Tab of the Inspect Window.

0x4007:
I'm mobile these days so I don't want to hold up the review somebody else please check on their computer for me.

.github/workflows/deploy.yml
```diff
@@ -18,12 +18,33 @@ jobs:
         with:
           repository: ${{ github.repository }}
           production_branch: ${{ github.event.repository.default_branch }}
-          build_artifact_name: "static"
-          output_directory: "static"
+          build_artifact_name: "full-stack-app"
+          output_directory: "full-stack-app"
```
0x4007: Full stack app, is this right? Why was this change introduced?

functions/issue-scraper.ts
```diff
@@ -0,0 +1,351 @@
...
+  } catch (error) {
+    console.error("Error processing request:", error);
+    return new Response("Internal Server Error", {
+      headers: corsHeaders,
+      status: 500,
+    });
+  }
+}
+
+function markdownToPlainText(markdown: string | null): string | null {
```
0x4007, line 162: Why do you do this conversion?

functions/issue-scraper.ts
```diff
@@ -0,0 +1,351 @@
...
+    markdown: string;
+    plaintext: string;
+    embedding: string;
+    author_id: number;
+    payload: PayloadType;
+  }>
+): Promise<void> {
+  const { error } = await supabase.from("issues").upsert(issues);
+  if (error) {
+    throw new Error(`Error during batch upsert: ${error.message}`);
```
0x4007: Perhaps we can try an exponential backoff? I wonder if this is worth implementing to make this more robust although I imagine that errors are quite rare for this.

functions/issue-scraper.ts
```diff
@@ -0,0 +1,351 @@
...
+    const octokit = new Octokit(token ? { auth: token } : {});
+    const voyageClient = new VoyageAIClient({ apiKey: voyageApiKey });
+
+    const issues = await fetchUserIssuesBatch(octokit, username, timestamp);
+
+    const uniqueAuthors = Array.from(new Set(issues.map((issue) => issue.author?.login).filter((login): login is string => !!login)));
+
+    const authorIdMap = await batchFetchAuthorIds(octokit, uniqueAuthors);
+
+    const markdowns = issues.map((issue) => `${issue.body || ""} ${issue.title || ""}`);
```
0x4007: I think that titles can NOT be empty so this should throw an error.

functions/issue-scraper.ts
```diff
@@ -0,0 +1,351 @@
...
+        id: issue.id,
+        markdown: markdowns[index],
+        plaintext: plainTexts[index] ?? "",
+        embedding: JSON.stringify(embeddings[index] || Array(VECTOR_SIZE).fill(0)),
+        author_id: authorId,
+        payload: {
+          issue: {
+            nodeId: issue.id,
+            number: issue.number,
+            title: issue.title || "",
```
0x4007: This can not be empty

src/home/authentication.ts
```diff
@@ -4,6 +4,8 @@ import { GitHubUser } from "./github-types";
 import { trackReferralCode } from "./register-referral";
 import { displayGitHubUserInformation } from "./rendering/display-github-user-information";
 import { renderGitHubLoginButton } from "./rendering/render-github-login-button";
+import { startIssueScraper } from "./scraper/issue-scraper";
+// import { issueScraper } from "./scraper/issue-scraper";
```
0x4007:
```suggestion
```

src/home/authentication.ts
```diff
@@ -20,5 +22,9 @@ export async function authentication() {
   if (gitHubUser) {
     await trackReferralCode();
     await displayGitHubUserInformation(gitHubUser);
+    // <-- Issue Scraper here -->
+    // const supabase = getSupabase();
```
0x4007: You can clean this file up more.

src/home/scraper/issue-scraper.ts
```diff
@@ -0,0 +1,49 @@
...
+  const now = Date.now();
+
+  // Prepare the request payload
+  const requestBody: { authToken: string; timestamp?: number } = {
+    authToken: supabaseAuth.provider_token,
+  };
+
+  if (lastFetch) {
+    const lastFetchTimestamp = Number(lastFetch);
+    if (now - lastFetchTimestamp < 24 * 60 * 60 * 1000) {
```
0x4007, line 38: I wonder if there is a better way to implement this. For example, instead of by timing, we can do based on an event. 

Although I can't think of how it would be possible in outside organization contexts.

src/home/scraper/issue-scraper.ts
```diff
@@ -0,0 +1,49 @@
...
+      return JSON.stringify({
+        success: true,
+        message: "Skipping fetch - last fetch was less than 24 hours ago",
```
0x4007, line 41: The login timeout currently lasts for a day I think, meaning that for 99% of users, they wouldn't retrigger a scrape with a new login in less than 24 hours. Something to note. 

We can also change this timeout in the supabase with settings. Any remarks on what to change it to or leave as is?

functions/issue-scraper.ts
shiv810: The `issues` table has two columns, `plaintext` and `markdown`. I think it would be a good idea to store the raw data and process it on the fly instead.

functions/issue-scraper.ts
shiv810: It should be possible to implement, but if a request fails and we retry multiple times, the entire request will eventually be terminated once it hits the maximum CPU wall time.

src/home/scraper/issue-scraper.ts
shiv810: Currently, when issues are marked as completed, a flag is set in the local storage, but it is not cleared upon logout. When the user logs in again, if the flag is older than 24 hours, its value is used to retrieve any recently closed issues.

.github/workflows/deploy.yml
shiv810: It could be anything; it's just a directory name. Changed this back to the original value.

functions/issue-scraper.ts
0x4007: Retries should be handled from the client so no timeouts on the worker side should be relevant.

0x4007 (approved):

functions/issue-scraper.ts
```diff
@@ -0,0 +1,368 @@
...
+
+function markdownToPlainText(markdown: string | null): string | null {
+  if (!markdown) return markdown;
+  const md = markdownit() as MarkdownItWithPlainText;
+  md.use(plainTextPlugin);
+  md.render(markdown);
+  return md.plainText;
+}
+
+const SEARCH_ISSUES_QUERY = `
```
gentlementlegen: You can add `/* GraphQL */` to help IDEs parse and format GraphQL queries.
```suggestion
const SEARCH_ISSUES_QUERY = /* GraphQL */`
```

functions/issue-scraper.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+`;
+
+async function fetchUserIssuesBatch(octokit: InstanceType<typeof Octokit>, username: string, lastScraped?: number): Promise<IssueNode[]> {
+  const allIssues: IssueNode[] = [];
+  let hasNextPage = true;
+  let cursor: string | null = null;
+
+  // Construct the query with the lastScraped timestamp
+  const searchText = `assignee:${username} is:issue is:closed reason:completed ${lastScraped ? `closed:>${new Date(lastScraped).toISOString()}` : ""}`;
```
gentlementlegen: Does this work if the user has a private profile?

package.json
```diff
@@ -36,10 +36,14 @@
     "@octokit/request-error": "^6.1.0",
     "@octokit/rest": "^20.0.2",
     "@supabase/supabase-js": "^2.39.0",
+    "@types/markdown-it": "^14.1.2",
```
gentlementlegen: Should be in `devDependencies`

yarn.lock
gentlementlegen, line 1: Shouldn't this be deleted since you used `bun`?

functions/issue-scraper.ts
shiv810: Yes, this would work with private profile, as we are querying with user's GitHub token.

yarn.lock
shiv810: I'm not entirely sure about this. @zugdev mentioned that Bun is related to Cloudflare, while Yarn is the main one.So, I've reverted the `bun.lockb` to match the repo version and added the dependencies to the `yarn.lock`.

gentlementlegen: I suppose it's okay to use either because the frontend doesn't rely on plugins or the kernel, but having both lockfiles seems unnecessary, I would suggest removing one or the other 😄

functions/issue-scraper.ts
shiv810: Should the retries be handled in this task? I think the error for this function should be very rare.

0x4007: That's what I would implement if I was working on this code yeah

shiv810: Added, a exponential backoff, the function endpoint forwards the x-rate-limit attributes if present in the header from the request, and the client retries based on that.

src/home/scraper/issue-scraper.ts
shiv810: I think using timing would be the best approach, as it would allow us to selectively identify closed issues after a certain point. This way, we can be sure that we’ve already scraped everything up until the timestamp provided.

shiv810:
@gentlementlegen @0x4007  Is there anything pending for this issue ?

gentlementlegen:
Let me give it a try locally, to get my own QA and should be ok for me.

shiv810:
> Let me give it a try locally, to get my own QA and should be ok for me.

you would need to add these three env variables  in cloudflare dashbaord/locally
Screenshot 2025-01-20 at 11 23 46 PM

In this case, it would be the supabase db for the text-embedding repo.

gentlementlegen:
Could you fix the Cypress as well?

shiv810:
> Could you fix the Cypress as well?

I can, but I would need to refactor the cypress tests. I think they're broken, as per this [issue](https://github.com/ubiquity/work.ubq.fi/issues/109).

0x4007 (approved):
I've had inconsistent problems with cypress so I'm inclined to let it slide.

functions/issue-scraper.ts
```diff
@@ -0,0 +1,464 @@
...
+        })),
+        storageFailed: storageFailed,
+      },
+      null,
+      2
+    );
+  } catch (error) {
+    console.error("Error in issueScraper:", error);
+    
+    if ((error as ApiError).source) {
```
gentlementlegen: Wouldn't `'source' in error` be safer? or `instance of ApiError` ?

gentlementlegen:
And could you add a `.dev.vars.example` for good measure? And probably a `.dev.vars` in the `.gitignore`.

README.md
```diff
@@ -14,6 +14,12 @@ yarn start
 open http://localhost:8080
 ```
 
+## Setup Scraper Function
+
+For local development of the scraper function, you need to set up a `.dev.vars` file based on the `.dev.vars.example` file.
+
+For deplotment, you would need to add the secrets to the Cloudflare Dashboard, for the worker.
```
gentlementlegen:
```suggestion
For deployment, you would need to add the secrets to the Cloudflare Dashboard, for the worker.
```

gentlementlegen (approved):

shiv810:
@zugdev @0x4007 or anyone with write access please merge this

PR #195, opened 2025-01-24
feat: avatar from github map

### LINKED ISSUE #194: Clear Cache Crash
image

I cleared my cache to get the latest organization avatars and now I can't get the app to work. 

We should also have a system to check the avatar and update it if it goes out of date once in awhile

zugdev:
Resolves #194 

You can QA this by opening console, indexedDb, GitHubAvatars, imageStore, then check the timestamps. Once 15 minutes have passed it should re-cache so timestamp should change.

image


src/home/fetch-github/fetch-avatar.ts
```diff
@@ -9,6 +9,12 @@ import { taskManager } from "../home";
 // Map to track ongoing avatar fetches
 const pendingFetches: Map<string, Promise<Blob | void>> = new Map();
 
+export async function fetchPartnerAvatars(): Promise<GitHubIssue[]> {
+  const response = await fetch("https://raw.githubusercontent.com/ubiquity/devpool-directory/__STORAGE__/devpool-partner-avatars.json");
```
0x4007, line 9: I have mixed feelings on your approach. The normal operating procedure is to create a pull request and link it to this task. Editing a production system is highly risky and prone to breakage. In the future you should work off of your own forks for a proof of concept. 

https://github.com/ubiquity/devpool-directory/commit/da89c2be5acc04877fce55dcf1b0a9a8c28854a6

Mixed feelings, though, because I could see myself doing the same thing given that we're the only users of this directory at the moment.

zugdev: mb, I only did that cause the change was independent from the rest of the system. so it was self contained. even if it failed it wouldnt compromise any other section, but I agree its dumb

zugdev:
Alright so in devpool-directory backend im fetching all orgs avatar_urls and storing in a map. Then once we load the page we also load the map. This way we dont even store images, its all URLs, we dont even need cache.

https://github.com/ubiquity/devpool-directory/blob/__STORAGE__/devpool-partner-avatars.json

image

0x4007:
You're missing an avatar
```json
[
    {
        "ownerName": "ubiquity",
        "avatar_url": "https://avatars.githubusercontent.com/u/76412717?v=4"
    },
    {
        "ownerName": "ubiquity-os-marketplace",
        "avatar_url": "https://avatars.githubusercontent.com/u/182627615?v=4"
    },
    {
        "ownerName": "0x4007"
    },
    {
        "ownerName": "ubiquity-os",
        "avatar_url": "https://avatars.githubusercontent.com/u/160213852?v=4"
    }
]
```

zugdev:
fixed

0x4007:
> fixed

In the future I suggest adding a link in the commit message. For example:

```
fix: add personal org avatar

https://github.com/ubiquity/work.ubq.fi/pull/195#issuecomment-2614124299
```

Then I can conveniently see the linkback in the timeline here and view the changes. 

These details help a ton with auditing the work later on.

src/home/fetch-github/fetch-avatar.ts
```diff
@@ -1,116 +1,33 @@
...
-import { organizationImageCache } from "./fetch-issues-full";
-import { GitHubIssue } from "../github-types";
-import { taskManager } from "../home";
+export const ubiquityAvatarUrl = "https://avatars.githubusercontent.com/u/76412717?v=4";
 
-// Map to track ongoing avatar fetches
-const pendingFetches: Map<string, Promise<Blob | void>> = new Map();
+export type OrgNameAndAvatarUrl = {
+  ownerName: string;
+  avatar_url?: string;
```
0x4007: I feel like its better practice to not make property name optional, but instead to initialize it as `null` if it is being set later in the program.

zugdev: good catch it doesnt even need to be nullable

0x4007 (approved):

0x4007:
Any luck with cypress? I would try on a new branch to fix because I always have problems with it

PR #202, opened 2025-02-13
feat: open issue in github in body absense and fix logo border-radius

zugdev:


0x4007:
The corner radius doesn't look like it matches. You normally shouldn't merge until approval. But good to see the link fix when they have no spec

zugdev:
> The corner radius doesn't look like it matches. You normally shouldn't merge until approval. But good to see the link fix when they have no spec

I evaluated it to be best to rapidly fix it for our new partners, I've opened a new PR to fix the logo https://github.com/ubiquity/work.ubq.fi/pull/203

PR #211, opened 2025-03-26
chore: update configuration

gentlementlegen:
This PR updates the `publicAccessControl` setting in the `daemon-pricing` configuration to `false`.


.github/.ubiquity-os.config.yml
```diff
@@ -1,67 +1,74 @@
-zeroTrue: &zeroTrue
+zeroTrue: &a3
```
0x4007: What is this new `a` schema?

gentlementlegen: The same as before just got shortened when generated

PR #212, opened 2025-04-02
feat: add org, repo and org/repo filters

### LINKED ISSUE #179: Search Adjustments
I was searching for https://devpool.directory/?search=top+up "top up" and its actually in the title of the task I was searching for. We need to increase the weight of the following:

1. task title
2. repository name
3. organization name

I think it could be interesting to prefix special operators to hone in for example imagine typing `@ubiquity/business-development` and then the UI would understand that you want to search only in the scope of that repository. This can be a shorthand hack instead of changing the url to `devpool.directory/ubiquity` but it would also make it clear that we are searching within a repository. Otherwise I suppose we can see if the search begins with a `/` and then it can signal that you are targeting the repository. 

In this case we can search `/business-development` and it would isolate the tasks from that repository. 

In conclusion we should 1. increase the weight of matching task titles and 2. we should allow these fine grained search filters by parsing the symbols `@` and `/` in the first character of the search.

ishowvel:
Resolves #179

### DIFF
--- a/src/home/sorting/filter-issues-by-search.ts
+++ b/src/home/sorting/filter-issues-by-search.ts
@@ -9,6 +9,7 @@ interface ParseResult {
 
 export function filterIssuesBySearch(filterText: string) {
   const { repoFilter, orgFilter, filteredSearchText } = parseSearchQuery(filterText);
+  console.error(repoFilter, orgFilter)
   const searchResults = taskManager.issueSearcher.search(filteredSearchText, orgFilter, repoFilter);
   // Create the new GithubIssue[] array based on the ranking in the searchResults
   const sortedIssues = Array.from(searchResults.entries())
--- a/src/home/sorting/filter-issues-by-search.ts
+++ b/src/home/sorting/filter-issues-by-search.ts
@@ -1,26 +1,16 @@
 import { GitHubIssue } from "../github-types";
 import { taskManager } from "../home";
 
+interface ParseResult {
+  orgFilter: string | undefined;
+  repoFilter: string | undefined;
+  filteredSearchText: string;
+}
+
 export function filterIssuesBySearch(filterText: string) {
-  let searchText = filterText
-  let orgFilter: string | undefined = undefined;
-  let repoFilter: string | undefined = undefined;
-
-  const pattern = /^(?:@(?<org>[^/\s]+)(?:\/(?<repo>[^\s]+))?|\/(?<repoOnly>[^\s]+))/;
-  const match = searchText.match(pattern);
-
-  if (match) {
-    if (match.groups?.org) {
-      orgFilter = match.groups.org;
-      repoFilter = match.groups.repo; 
-    } else if (match.groups?.repoOnly) {
-      repoFilter = match.groups.repoOnly;
-    }
-    // Remove the matched filter segment from the beginning of the search text.
-    searchText = searchText.slice(match[0].length).trim();
-  }
-
-  const searchResults = taskManager.issueSearcher.search(searchText, orgFilter, repoFilter);
+  const { repoFilter, orgFilter, filteredSearchText } = parseSearchQuery(filterText);
+  console.error(parseSearchQuery(filterText));
+  const searchResults = taskManager.issueSearcher.search(filteredSearchText, orgFilter, repoFilter);
   // Create the new GithubIssue[] array based on the ranking in the searchResults
   const sortedIssues = Array.from(searchResults.entries())
     .filter(([, result]) => result.score > 0)
@@ -29,4 +19,39 @@ export function filterIssuesBySearch(filterText: string) {
     .filter((issue): issue is GitHubIssue => issue !== undefined);
 
   return sortedIssues;
+}
+
+export function parseSearchQuery(searchText: string): ParseResult {
+  const input = searchText.trim();
+  const lower = input.toLowerCase();
+
+  if (!lower.startsWith("/") && !lower.startsWith("@")) {
+    return {
+      filteredSearchText: lower,
+      orgFilter: undefined,
+      repoFilter: undefined,
+    };
+  }
+  
+  const pattern = /^(?:@([^/\s]+)(?:\/([^/\s]+))?|\/([^/\s]+))/;
+  const match = pattern.exec(lower);
+
+  // If regex fails (i.e. it was just "/" or just "@"), return empty search
+  if (!match) {
+    return {
+      orgFilter: undefined,
+      repoFilter: undefined,
+      filteredSearchText: "",
+    };
+  }
+
+  const orgFilter = match[1];
+  const repoFilter = match[2] ?? match[3];
+  const rest = lower.slice(match[0].length).trim();
+
+  return {
+    orgFilter,
+    repoFilter,
+    filteredSearchText: rest,
+  };
 }
--- a/src/home/sorting/filter-issues-by-search.ts
+++ b/src/home/sorting/filter-issues-by-search.ts
@@ -9,7 +9,6 @@ interface ParseResult {
 
 export function filterIssuesBySearch(filterText: string) {
   const { repoFilter, orgFilter, filteredSearchText } = parseSearchQuery(filterText);
-  console.error(parseSearchQuery(filterText));
   const searchResults = taskManager.issueSearcher.search(filteredSearchText, orgFilter, repoFilter);
   // Create the new GithubIssue[] array based on the ranking in the searchResults
   const sortedIssues = Array.from(searchResults.entries())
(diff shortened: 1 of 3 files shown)


ishowvel:
@0x4007 for increasing title strength i dont understand why thats necessary, i searched issues using their title and the respective issue always came at the top of the search result

ishowvel:
> @0x4007 for increasing title strength i dont understand why thats necessary, i searched issues using their title and the respective issue always came at the top of the search result

@0x4007

ishowvel:
> ```diff
> ! Error: Error: 402 Insufficient credits. Add more using https://openrouter.ai/settings/credits
> ```

@0x4007

zugdev:
hey, the @org/repo regex doesn't match for me:

image

src/home/sorting/filter-issues-by-search.ts
```diff
@@ -2,12 +2,31 @@ import { GitHubIssue } from "../github-types";
 import { taskManager } from "../home";
 
 export function filterIssuesBySearch(filterText: string) {
-  const searchResults = taskManager.issueSearcher.search(filterText);
-  //Create the new GithubIssue[] array based on the ranking in the searchResults
+  let searchText = filterText
+  let orgFilter: string | undefined = undefined;
```
zugdev, line 6: I rather use null instead of undefined

zugdev:
the regex works, it's matching right for org and repo

image

however in search():

```
issue.repository?.owner.login !== orgFilter || repoFilter && issue.repository?.name !== repoFilter
``` 
both can be undefined because of ?, so it returns false for all. use a known existing property like `repository_url` instead

ishowvel:
image
@zugdev

ishowvel:
can you please approve changes so i can start on the user onboarding flow

ishowvel:
@zugdev

zugdev:
> @zugdev 

I've pointed out a problem, can you elaborate on it? Important before approving.

ishowvel:
> > @zugdev 
> 
> I've pointed out a problem, can you elaborate on it? Important before approving.

The problem has been fixed, I am now using repo url to get repo owner and repo name, @ and / search yields issues under some org and issues with some repo name respectively

ishowvel:
> image
> @zugdev 

Qa

ishowvel:
The commit which fixed the problem https://github.com/ubiquity/work.ubq.fi/commit/af21f1ac272b422d5ad7d89cf594df321c0e3d1d

zugdev:
`@org` matching definitely works!

image

`/repo` matching definitely works!

image

`@org/repo` doesn't work.
image


Problems I noticed:

1. Currently your org matching only works for exact matches, so if I pass in "@ubiqui" it correctly matches nothing. 

image

Perhaps when there is no exact match we can either:
- do partial matching: including any org starting with "ubiqui"
- return normal search results

This is motivated by UX, since it feels weird initially having a list of tasks, you type @ and everything is gone:
image

2. After you do org filtering if you clear your search you get no more normal results, in fact once you go into org filtering the only way out is refreshing:

image

Suggested pseudocode: 
```
let matches = []
if searchString[0] === "@"{
     matches = match org name and potentially repo name
} elseIf searchString[0] === "/"{
     matches = match repo name
}

return matches ? matches : normal search (currently implemented)
```

zugdev:
Almost there @ishowvel, I promise to quickly review this from now on. Tag me when you are done.

ishowvel:
partial `@org` match
image

partial `/repo` match
image

partial `@org/repo` match
image

exact `@org` match
image

exact `/repo` match 
image

exact `@org/repo` match
image

empty filter
image

> After you do org filtering if you clear your search you get no more normal results, in fact once you go into org filtering the only way out is refreshing:

image

ishowvel:
@zugdev

ishowvel:
I suggest switching from yarn to bun in the build workflow

ishowvel:
I suggest switching from yarn to bun in workflows

zugdev:
Nice QA, thanks - it's very helpful. I found two bugs:

1. Normal search no longer works when the regex is not a match with @ or /

Uploading Screen Recording 2025-04-27 at 22.09.51.mov…

2. When you toggle between proposals and directory the regex stops matching

Uploading Screen Recording 2025-04-27 at 22.11.05.mov…

> I suggest switching from yarn to bun in workflows

Yes we should migrate entirely, I'll spec it.

zugdev:
Nice QA, thanks - it's very helpful. I found two bugs:

1. Normal search no longer works when the regex is not a match with @ or /

https://github.com/user-attachments/assets/caadf833-027b-4f4a-9780-b8b1055932cf

2. When you toggle between proposals and directory the regex stops matching

https://github.com/user-attachments/assets/dd100430-0f0d-424c-b3de-783c362f943f

> I suggest switching from yarn to bun in workflows

Yes we should migrate entirely, I'll spec it.

ishowvel:
> partial `@org` match image
> 
> partial `/repo` match image
> 
> partial `@org/repo` match image
> 
> exact `@org` match image
> 
> exact `/repo` match image
> 
> exact `@org/repo` match image
> 
> empty filter image
> 
> > After you do org filtering if you clear your search you get no more normal results, in fact once you go into org filtering the only way out is refreshing:
> 
> image

i have tested all these cases plus switching between proposals and directory and fixed normal searching

src/home/sorting/sorting-manager.ts
```diff
@@ -5,6 +5,7 @@ import { Sorting } from "./generate-sorting-buttons";
 import { isFilteringAvailableIssues, swapAvailabilityFilter } from "../fetch-github/fetch-and-display-previews";
 
 export class SortingManager {
+  private static _instances: SortingManager[] = [];
```
zugdev, line 8: what does this do?

zugdev:
This works for the previously stated purposes - thanks! There is still one last glitch before I can approve it, if I change a middle part of the search string, search breaks.

https://github.com/user-attachments/assets/cb96277f-a51a-4316-a69f-01b0eed8a5c0

ishowvel:
> This works for the previously stated purposes - thanks! There is still one last glitch before I can approve it, if I change a middle part of the search string, search breaks.
> 
> https://github.com/user-attachments/assets/cb96277f-a51a-4316-a69f-01b0eed8a5c0
> 

I thought we didn't need to implement partial org matching for @org/repo, will implement tomorrow

src/home/sorting/sorting-manager.ts
ishowvel: this is to sync the top and bottom sorting managers, their text boxes aren't synced which causes incorrect issue searching when switching between directory and proposal

You can see this yourself after commenting out the logic for this, typing something in the top text box and then checking the bottom text boxes value

zugdev:
> > This works for the previously stated purposes - thanks! There is still one last glitch before I can approve it, if I change a middle part of the search string, search breaks.
> > 
> >   
> >     
> >     
> > 
> >     Screen.Recording.2025-04-29.at.11.53.32.mov
> 
> I thought we didn't need to implement partial org matching for @org/repo, will implement tomorrow

That's not exactly the problem. If I type "@ubiquity-os/plugins-wishlist", then delete some of the @ part: "@.../plugins-wishlist" and add it again "@ubiquity-os/plugins-wishlist" it breaks. But I suppose partial match could solve it, great work so far.

ishowvel:
> That's not exactly the problem. If I type "@ubiquity-os/plugins-wishlist", then delete some of the @ part: "@.../plugins-wishlist" and add it again "@ubiquity-os/plugins-wishlist" it breaks. But I suppose partial match could solve it, great work so far.

will do

ishowvel:
Why did this close

ishowvel:
@zugdev I am really sorry I'll make a new pr

ishowvel:
@zugdev should work now

zugdev:
> @zugdev should work now

Will test soon, thanks

zugdev (approved):
This works well

ishowvel:
@0x4007 can you review this

ishowvel:
@zugdev does this need pavlovciks reviews or can this be merged without it

zugdev:
we need him, he will eventually check it out, don't worry

ishowvel:
> This works well

@0x4007 can you review this pr, it already has a approved review and is just waiting for yours to be merged

ishowvel:
@0x4007 waiting for review

ishowvel:
> @ishowvel, this task has been idle for a while. Please provide an update on your progress.
> 

Completely forgot about this due to inactivity in ubiquity, will qa today

ishowvel:
> @ishowvel, this task has been idle for a while. Please provide an update on your progress.
> 

@zugdev what should I do the bot keeps pinging me 😭

zugdev:
> > @ishowvel, this task has been idle for a while. Please provide an update on your progress.
> > 
> 
> @zugdev what should I do the bot keeps pinging me 😭

You just gotta wait, you'll be reassigned in case you are automatically unassigned. No need to reply it.

ishowvel:
@zugdev will this be reviewed by pavlovcik when bounties are re enabled for external contributors?

ishowvel:
@0x4007 can you review this

0x4007 (approved):

PR #216, opened 2025-08-02
Display organization in preview modal

### LINKED ISSUE #207: Display Organization in Modal
image

I was just going through some proposals to clean them out and I realized that we actually don't display the organization name in this UI so we should definitely do that. 

It looks like we have plenty of extra screen real estate on the bottom left so I suppose we can place their or organization, avatar and style it in the same way we have it in the list view. 

Obviously this must also look good on desktop.

gentlementlegen:
Resolves #207
ScreenRecording2025-08-04at20 11 09-ezgif com-resize


0x4007:
I guess continuous deployments are disabled. 

The letter spacing does not match how the normal header looks on your video. Match the styles please.

gentlementlegen:
@0x4007 The deployments seem to fail: https://github.com/ubiquity/work.ubq.fi/actions/runs/16730735212 I don't know if it is because I am linking this from my personal account (as a fork)

I did not change any styling, this is literally how it displays without my changes, not sure what you mean?

gentlementlegen:
@0x4007 I fixed the padding, didn't realize this class name was already in use, should be good now.

0x4007 (approved):

src/home/rendering/render-org-header.ts
```diff
@@ -13,10 +14,56 @@ export function renderOrgHeaderLabel(orgName: string): void {
...
+
+  const brandingDiv = document.getElementById("branding");
+  if (!brandingDiv) return;
+
+  const pathSegments = window.location.pathname.split("/").filter(Boolean);
+  const urlOrgName = pathSegments.length > 0 ? pathSegments[0] : null;
+
+  const currentLogo = brandingDiv.querySelector("img#logo") ?? brandingDiv.querySelector("svg#logo");
+  if (!urlOrgName && currentLogo) {
+    currentLogo.replaceWith(createUbiquitySvg());
```
0x4007, line 65: It would be way nicer if they didn't have to load in every time. You should keep the contents in memory as a base64 encoded string or something. Because every time I switch, the logo renders from top to bottom slowly.

gentlementlegen: I figured the website was handling image caching but maybe not then, opening a task for this.

0x4007: I just realized it could be because I had my dev tools open with cache disabled.

gentlementlegen:
@0x4007 I guess the lock files should not count towards you review and should not be claimed

0x4007:
> [!TIP]
> Hello, world!

PR #218, opened 2025-10-12
Integrate new storage artifacts + proposals support

### LINKED ISSUE #219: Upgrade UI to support devpool-directory v2
After doing a bunch of changes to the backend, we've changed the artifacts format. Fixes need to be made to support the Issues view and the Proposals view.

0x4007:
Resolves #219

This PR updates the UI to the new DevPool Directory storage format and restores proposals view.

Follow-ups:
- Optionally move to GitHub Contents API + ETags via `summary.json` for polling as per storage README.

References:
- Storage guide: https://github.com/devpool-directory/devpool-directory/blob/__STORAGE__/README_STORAGE.md


0x4007:
Ready for re-review — v2 storage integration + ‘Price:’ only

Summary of updates:
- Deprecate ‘Pricing:’ label; only ‘Price: ’ is recognized in filtering, rendering, and price sorting.
- Upgrade Unassigned/All Issues toggle for v2 backend using mirror-state:
  - Integrate mirror-state.json (assigned, assignees) into issue mapping
  - Availability filter prefers assigned flag with fallback to assignee/assignees
  - UI label renamed to ‘Available’ ↔ ‘All Issues’
- Refactors from prior review:
  - Add isNamedLabel guard and simplify priority parsing
  - Replace placeholder assignee with explicit object; remove double assertion
  - Typed availability filter
- Resolved all handled review threads for the touched files
- Build and ESLint pass

Refs:
- __STORAGE__ integration guide: https://github.com/devpool-directory/devpool-directory/blob/__STORAGE__/README_STORAGE.md
- Commits: 05ea9a4, d69a9ea, 747ee73, 9496d36

src/home/sorting/sort-issues-by-price.ts
```diff
@@ -10,7 +11,7 @@ export function sortIssuesByPrice(issues: GitHubIssue[]) {
 }
 
 function getPriceFromLabel(label: string | { name?: string }) {
-  const text = typeof label === "string" ? label : label?.name || "";
+  const text = getLabelText(label as unknown as GitHubIssue["labels"][number]);
```
gentlementlegen: What is the point of that cast since the type of label is `(label: string | { name?: string })`?

src/home/rendering/render-github-issues.ts
```diff
@@ -9,6 +9,8 @@ import { bottomBar, bottomBarClearLabels, closeModal, modal, modalBodyInner, tit
 import { setupKeyboardNavigation } from "./setup-keyboard-navigation";
 import { waitForElement } from "./utils";
 
+const LABEL_PREFIX_REGEX = /^(Price: |Time: |Priority: )/;
```
gentlementlegen, line 13: `Priority` is technically not enforced, the configuration for this one can be changed

0x4007: Let's revert to only support the full notation for now. `Priority: `

gentlementlegen:
We can keep the STORAGE branch logic for v1, but wouldn't it be more appropriate to use Deno KV for this eventually?

Also, Cypress was removed but the workflow is still present in the CI. Otherwise, proposals and the nomal directory are displayed properly, although they contain closed issues due to the 4 days old sync.

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -54,11 +54,12 @@ viewToggle.addEventListener("click", () => {
...
   return (issue: GitHubIssue) => {
-    if (!issue?.labels) return false;
-
-    const hasPriceLabel = issue.labels.some((label) => {
-      if (typeof label === "string") return false;
-      return label.name?.startsWith("Price: ") || label.name?.startsWith("Price: ");
+    const labels = Array.isArray(issue?.labels) ? issue.labels : [];
+    const hasPriceLabel = labels.some((label) => {
+      if (typeof label === "string") {
+        return /^(Price:)\s*/.test(label);
```
gentlementlegen: could be simplified to something like

```typescript
return /^(Price:)\s*/.test(typeof label === "string" ? label : label.name);
```

src/home/fetch-github/fetch-issues-full.ts
```diff
@@ -3,13 +3,94 @@ import { GitHubIssue } from "../github-types";
...
+  directory_issue_url?: string;
+};
+
+// Generate a stable 53-bit hash to minimize collisions for node_id
+// Based on a dual-DJB2 accumulator producing up to 53 bits.
+// The magic numbers 5381 and 52711 are used as initial values for the hash accumulators:
+//   - 5381 is the traditional starting value for the DJB2 hash function, chosen by Daniel J. Bernstein for its good distribution properties.
+//   - 52711 is a large prime number used as a second accumulator to extend the bit width and further reduce hash collisions.
+// This dual-accumulator approach helps produce a more stable and unique hash for string inputs.
+function hashStringToNumber(input: string): number {
```
gentlementlegen, line 39: What is the purpose of this? According to the docs, `node_id` is globally unique.

src/home/rendering/render-github-issues.ts
```diff
@@ -95,49 +97,36 @@ function setUpIssueElement(issueElement: HTMLDivElement, task: GitHubIssue, orga
...
+  const acc = { labels: [] as { order: number; label: string }[], otherLabels: [] as string[] };
 
-      // check if label.name exists
-      if (!label.name) {
-        return {
-          labels: [],
-          otherLabels: [],
-        };
-      }
+  for (const label of task.labels || []) {
```
gentlementlegen, line 105:
```suggestion
  for (const label of task.labels) {
```

src/home/rendering/render-github-issues.ts
```diff
@@ -95,49 +97,36 @@ function setUpIssueElement(issueElement: HTMLDivElement, task: GitHubIssue, orga
...
+    const name = typeof label === "string" ? label : label?.name || "";
```
gentlementlegen:
```suggestion
    const name = typeof label === "string" ? label : label?.name;
```

0x4007:
> We can keep the STORAGE branch logic for v1, but wouldn't it be more appropriate to use Deno KV for this eventually?

Works perfectly fine as commits. The entire thing runs using GitHub only. 

> 
> Also, Cypress was removed but the workflow is still present in the CI. Otherwise, proposals and the nomal directory are displayed properly, although they contain closed issues due to the 4 days old sync.

I forgot to set the cron job but its intended to run automatically, maybe hourly.

src/home/search/search-scorer.ts
```diff
@@ -72,16 +72,15 @@ export class SearchScorer {
...
-            const labelName = label.name.toLowerCase();
-            if (labelName.includes(term)) {
-              matchDetails.labelMatches.push(label.name);
-              // Apply exponential boost for label matches at word start
-              if (labelName.startsWith(term)) {
-                score += 0.8;
-              } else {
-                score += 0.5;
-              }
+          const labelName = (typeof label === "object" && (label as any).name ? (label as any).name : typeof label === "string" ? label : "").toLowerCase();
```
gentlementlegen: `any` casts should no be needed here, label is already typed.

src/home/sorting/filter-issues-by-availability.ts
```diff
@@ -2,7 +2,11 @@ import { GitHubIssue } from "../github-types";
 
 export function filterIssuesByAvailability(issues: GitHubIssue[]) {
   return issues.filter((issue) => {
+    // v2 backend: prefer explicit `assigned` from mirror-state
+    if (issue.assigned === true) return false;
+    // Fallback: Treat either a non-null assignee or non-empty assignees array as unavailable
     if (issue.assignee) return false;
+    if (Array.isArray(issue.assignees) && issue.assignees.length > 0) return false;
```
gentlementlegen:
```suggestion
    return !(Array.isArray(issue.assignees) && issue.assignees.length > 0);
```

src/home/sorting/label-utils.ts
```diff
@@ -0,0 +1,5 @@
+import { GitHubLabel } from "../github-types";
+
+export function getLabelText(label: GitHubLabel): string {
+  return typeof label === "string" ? label : label?.name || "";
```
gentlementlegen, line 4: I have seen this logic many times before, either use this function everywhere or delete it

src/home/sorting/sort-issues-by-time.ts
```diff
@@ -1,10 +1,15 @@
 import { GitHubIssue } from "../github-types";
 import { calculateTimeLabelValue } from "./calculate-time-label-value";
+import { getLabelText } from "./label-utils";
 
 export function sortIssuesByTime(tasks: GitHubIssue[]) {
   return tasks.sort((a, b) => {
-    const aTimeValue = a.labels.reduce((acc, label) => acc + (typeof label === "object" && label?.name ? calculateTimeLabelValue(label.name) : 0), 0);
-    const bTimeValue = b.labels.reduce((acc, label) => acc + (typeof label === "object" && label?.name ? calculateTimeLabelValue(label.name) : 0), 0);
-    return bTimeValue - aTimeValue;
+    const sumTime = (labels: GitHubIssue["labels"]) =>
```
gentlementlegen: might want to use function style instead of anonymous functions here?

src/home/issues-search.ts
```diff
@@ -233,7 +233,8 @@ export class IssueSearch {
 
     const title = issue.title;
     const body = removeUrls(issue.body || "");
-    const labels = issue.labels?.map((l) => (typeof l === "object" && l.name ? l.name : "")).join(" ") || "";
+    const labels =
+      issue.labels?.map((l) => (typeof l === "object" && (l as any).name ? (l as any).name : typeof l === "string" ? l : "")).join(" ") || "";
```
gentlementlegen: any should not be required

static/style/style.css
```diff
@@ -347,9 +347,9 @@
     max-height: 48px;
   }
   #authenticated > img {
-    height: 35px;
+
```
gentlementlegen, line 350:
```suggestion
```

package.json
```diff
@@ -20,9 +20,7 @@
     "knip": "knip --config .github/knip.ts",
     "knip-ci": "knip --no-exit-code --reporter json --config .github/knip.ts",
     "prepare": "husky install",
-    "test": "jest --setupFiles dotenv/config --coverage",
-    "cy:open": "cypress open",
-    "cy:run": "cypress run"
+    "test": "jest --setupFiles dotenv/config --coverage"
```
gentlementlegen: this can also be removed there is no testing suite at all now

build/esbuild-build.ts
```diff
@@ -9,9 +9,12 @@ const typescriptEntries = ["src/home/home.ts"];
 const cssEntries = ["static/style/style.css"];
 const entries = [...typescriptEntries, ...cssEntries, "static/manifest.json", "static/favicon.svg", "static/icon-512x512.png"];
 
+const isProd = (process.env.NODE_ENV || "development") === "production";
```
gentlementlegen:
```suggestion
const isProd = process.env.NODE_ENV === "production";
```

build/esbuild-build.ts
```diff
@@ -9,9 +9,12 @@ const typescriptEntries = ["src/home/home.ts"];
...
+
 export const esBuildContext: esbuild.BuildOptions = {
   plugins: [invertColors, pwaManifest],
-  sourcemap: true,
+  // Use inline sourcemaps in dev to avoid network/map caching issues; linked maps in prod
+  sourcemap: isProd ? true : "inline",
```
gentlementlegen, line 12: Why would we want to slow down production with `sourcemap`? Shouldn't it be the other way around?

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -54,12 +55,8 @@ viewToggle.addEventListener("click", () => {
 
 function getProposalsOnlyFilter(getProposals: boolean) {
   return (issue: GitHubIssue) => {
-    if (!issue?.labels) return false;
-
-    const hasPriceLabel = issue.labels.some((label) => {
-      if (typeof label === "string") return false;
-      return label.name?.startsWith("Price: ") || label.name?.startsWith("Price: ");
-    });
+    const labels = Array.isArray(issue?.labels) ? issue.labels : [];
```
gentlementlegen: In what scenario this is not an array? The type indicates that it should be a `GitHubLabel[]`

src/home/fetch-github/fetch-issues-full.ts
```diff
@@ -3,13 +3,96 @@ import { GitHubIssue } from "../github-types";
...
+  // Combine to a positive 53-bit integer
+  const combined = (h1 >>> 0) * 0x1000 + ((h2 >>> 0) & 0x0fff);
+  return combined;
+}
+
+function mapStorageIssueToGitHubIssue(issue: StorageIssue, mirror?: MirrorStateEntry): GitHubIssue {
+  const repositoryUrl = `https://github.com/${issue.owner}/${issue.repo}`;
+  // Note: UI code expects a numeric `id` to use in DOM attributes and URL params.
+  // GitHub's `node_id` is globally unique but a string; we derive a stable numeric id from it.
+  const id = hashStringToNumber(issue.node_id);
```
gentlementlegen, line 62: I still don't understand why this has to be a number. I was checking the code and in `render-github-issues.ts` the check  is
```typescript
if (!existingIssueIds.has(task.id.toString())) {
```

So we transform back the string to number back to string. In the url params, params are always treated as a string as well. So instead of `?issue=126456` it could be `?issue=node_123456` and it would most likely work the same.

src/home/sorting/filter-issues-by-availability.ts
gentlementlegen:
```suggestion
    return !(Array.isArray(issue.assignees) && issue.assignees.length > 0);
```

src/home/sorting/sort-issues-by-priority.ts
```diff
@@ -1,15 +1,24 @@
...
 
   return issues.sort((a, b) => {
     function getPriority(issue: GitHubIssue) {
-      const priorityLabel = issue.labels.find(
-        (label): label is { name: string } => typeof label === "object" && "name" in label && typeof label.name === "string" && priorityRegex.test(label.name)
-      );
-      const match = priorityLabel?.name.match(priorityRegex);
-      return match ? parseInt(match[1], 10) : -1;
+      // Try object labels
+      const objectLabel = issue.labels.find((l): l is { name: string } => isNamedLabel(l) && priorityRegex.test(l.name));
```
gentlementlegen: Instead of making two loops, why not just one loop with
```typescript
const name = typeof label === 'string' ? label : isNamedLabel(label) ? label.name : '';
```

Or even better using `getLabelText` utility function directly

src/home/sorting/sorting-manager.ts
```diff
@@ -194,7 +194,7 @@ export class SortingManager {
 
     input.addEventListener("click", () => {
       swapAvailabilityFilter();
-      input.value = isFilteringAvailableIssues ? "Unassigned" : "All Issues";
+      input.value = isFilteringAvailableIssues ? "Available" : "All Issues";
```
gentlementlegen: Might be worth storing these in constants since they are reused

static/style/style.css
```diff
@@ -992,3 +993,9 @@
     display: none;
   }
 }
+
+#enable-private-issues-button {
+    border: none;
```
gentlementlegen, line 998: Spacing seems wrong

gentlementlegen:
`cypress-testing.yml` has not been deleted still.

0x4007:
#221

PR #221, opened 2025-10-17
chore(deno): migrate off Wrangler to pure Deno

### LINKED ISSUE #219: Upgrade UI to support devpool-directory v2

### LINKED ISSUE #220: Migrate from Cloudflare Wrangler to pure Deno server
Replace Wrangler dev setup with a Deno server and port Pages Functions to Deno handlers.

- Remove Wrangler/Cloudflare config and dependencies
- Add Deno.serve static server and API routes
- Port /issue-scraper and /referral-manager to Deno
- Use Deno KV for referrals
- Update scripts and docs

Endpoints remain the same for the frontend.

0x4007:
Resolves https://github.com/ubiquity/work.ubq.fi/issues/220

- Replace Wrangler dev with Deno.serve
- Port Pages Functions to Deno handlers
- Use Deno KV for referrals
- Remove CF config/deps; update docs and scripts

Fixes #219
Fixes #220


0x4007:
Marked all review threads as resolved since the code changes are now in. I’ll monitor the PR for any new comments or checks and follow up promptly.

package.json
```diff
@@ -9,20 +9,16 @@
     "node": ">=20.10.0"
   },
   "scripts": {
-    "start": "run-p start-wrangler watch",
-    "start-wrangler": "npx wrangler pages dev static --port 8080",
-    "watch": "tsx build/esbuild-watch.ts",
-    "build": "tsx build/esbuild-build.ts",
+    "start": "deno task dev",
```
gentlementlegen: Any reason to use `task` for these? It should be used to trigger parallel scripts.

Also, this does not run due to `error: Module not found "file:///work.ubq.fi/functions/validators". Maybe add a '.ts' extension or run with --sloppy-imports
` because it is missing a flag, Did you test this?

deno.json
```diff
@@ -0,0 +1,21 @@
+{
+  "tasks": {
+    "serve": "deno run --allow-net --allow-read --allow-env server.ts",
+    "build": "deno run -A build/esbuild-build.ts",
+    "watch": "deno run -A build/esbuild-watch.ts",
+    "dev": "deno run -A scripts/dev.ts",
```
gentlementlegen, line 13: `--sloppy-imports` is needed for it to run properly.

deno.json
```diff
@@ -0,0 +1,21 @@
...
+    "useTabs": false,
+    "lineWidth": 120,
+    "indentWidth": 2
+  },
+  "lint": {
+    "rules": {
+      "tags": ["recommended"]
+    }
+  }
+}
```
gentlementlegen, line 42:
```suggestion
"unstable": ["sloppy-imports", "kv"]
```

functions/issue-scraper.ts
```diff
@@ -1,11 +1,11 @@
-import { Context } from "./types";
-import { SupabaseClient } from "@supabase/supabase-js";
-import { VoyageAIClient, VoyageAIError } from "voyageai";
-import { Octokit } from "@octokit/rest";
-import markdownit from "markdown-it";
-import plainTextPlugin from "markdown-it-plain-text";
+// Deno-compatible request handler for the issue scraper endpoint
+import { SupabaseClient } from "npm:@supabase/supabase-js";
```
gentlementlegen: When trying to run this I get

```
error: Uncaught SyntaxError: The requested module 'npm:@supabase/supabase-js' does not provide an export named 'SupabaseClient'
```

functions/issue-scraper.ts
```diff
@@ -109,53 +109,61 @@ interface GraphQlSearchResponse {
...
+                  source: "configuration",
+                  status: 503,
+                  retryAfter: 3600,
+                  message: `Service not configured. Missing: ${missing.join(", ")}`,
+                },
+              }),
+              { status: 503, headers: { ...corsHeaders, "Content-Type": "application/json", "Retry-After": "3600" } }
+            );
+          }
+          const supabase = new SupabaseClient(SUPABASE_URL, SUPABASE_KEY);
```
gentlementlegen: This does not exist. It should be
```suggestion
          const supabase = createClient(SUPABASE_URL, SUPABASE_KEY);
```

And it should have typings.

server.ts
```diff
@@ -0,0 +1,31 @@
+// Load environment variables from .env in development
+import "https://deno.land/std@0.224.0/dotenv/load.ts";
+import { serveDir } from "https://deno.land/std@0.224.0/http/file_server.ts";
+import { handleIssueScraper } from "./functions/issue-scraper.ts";
+import { handleReferralManager } from "./functions/referral-manager.ts";
+
+const port = Number(Deno.env.get("PORT")) || 8080;
+
+const kv = await Deno.openKv();
```
gentlementlegen, line 10: This can't work with the current flags. You need to enable Deno KV in the `deno.json` configuration.

scripts/dev.ts
```diff
@@ -0,0 +1,35 @@
+// Simple dev runner to start server and esbuild watch concurrently
+const serve = new Deno.Command(Deno.execPath(), {
+  args: ["run", "--allow-net", "--allow-read", "--allow-env", "server.ts"],
+  stdout: "inherit",
+  stderr: "inherit",
+});
+
+const watch = new Deno.Command(Deno.execPath(), {
+  args: ["run", "-A", "build/esbuild-watch.ts"],
```
gentlementlegen, line 14: Why do we even need `esbuild` since Deno is serving the project now? Deno has bundling options itself: https://docs.deno.com/runtime/reference/bundling/#basic-usage

gentlementlegen:
When running the server, it just returns 404 pages so far.

0x4007:
> When running the server, it just returns 404 pages so far.

Oh I just realized due to the pwa I have to unload it to truly test the ui I guess I'll test using curls from shell when editing again.

0x4007:
Resolved all outstanding review threads with the following updates:\n\n- deno.json: add --allow-write to serve; set unstable [kv, sloppy-imports].\n- scripts/dev.ts: add --allow-write to server spawn.\n- server.ts: SPA fallback; global OPTIONS preflight; dynamic imports for handlers.\n- validators: use GET /user destructure; drop misleading type.\n- github-types: set GitHubUserResponse to getAuthenticated.\n- issue-scraper: use createClient(SUPABASE_URL, SUPABASE_KEY).\n\nCommits: d166cae, 9ae3a4e.\n\nIf anything else is needed, happy to adjust.]}

deno.json
```diff
@@ -1,9 +1,9 @@
 {
   "tasks": {
-    "serve": "deno run --unstable-kv --allow-net --allow-read --allow-env server.ts",
+    "serve": "deno run --unstable-kv --allow-net --allow-read --allow-env --allow-write server.ts",
     "build": "deno run -A build/esbuild-build.ts",
     "watch": "deno run -A build/esbuild-watch.ts",
-    "dev": "deno run -A scripts/dev.ts",
+    "dev": "deno run --unstable-kv -A scripts/dev.ts",
```
gentlementlegen: No need to have it here since it was added within the "unstable" section as well

gentlementlegen:
Could you explain the reason to still use `esbuild` here?

scripts/dev.ts
```diff
@@ -0,0 +1,54 @@
+// Simple dev runner to start server and esbuild watch concurrently
+// Picks an available port automatically unless PORT is provided by the user.
+const desiredPort = Deno.env.get("PORT") ?? "0"; // 0 = auto-assign an open port
+
+const serve = new Deno.Command(Deno.execPath(), {
```
gentlementlegen, line 6: I do not understand why all of this is needed

server.ts
```diff
@@ -0,0 +1,116 @@
...
+  } catch (err) {
+    // If the port is in use, retry with an ephemeral port
+    const msg = (err && (err as Error).message) || "";
+    const isAddrInUse = msg.includes("AddrInUse") || msg.includes("address already in use");
+    if (isAddrInUse && portToUse !== 0) {
+      console.warn(`Port ${portToUse} in use; retrying on an open port...`);
+      Deno.serve(
+        {
+          port: 0,
+          onListen: async ({ hostname, port }) => {
```
gentlementlegen: This code seems duplicated from just above

src/home/fetch-github/fetch-and-display-previews.ts
```diff
@@ -31,9 +32,9 @@ export function swapAvailabilityFilter() {
 }
 
 // start at view based on URL
-export let isProposalOnlyViewer = new URLSearchParams(window.location.search).get("proposal") === "true";
+let isProposalOnlyViewer = new URLSearchParams(window.location.search).get("proposal") === "true";
```
gentlementlegen: The window variable was removed in Deno 2.  `globalThis` should be used instead.

src/home/getters/get-github-access-token.ts
```diff
@@ -46,7 +23,17 @@ export async function getGitHubAccessToken(): Promise<string | null> {
...
+export async function isMissingRepoScope(): Promise<boolean> {
+  const token = await getGitHubAccessToken();
+  if (!token) return false;
+  const octokit = new Octokit({ auth: token });
+  const { headers } = await octokit.request("HEAD /");
+  const scopes = headers?.["x-oauth-scopes"]?.split(", ");
+  return !scopes?.includes("repo");
+}
+
+function getGitHubUserName(): string | null {
```
gentlementlegen: this function seems to never be used

gentlementlegen:
It seems that external organization task items are broken. When clicking on the item, it redirects to the task instead of opening the popup with the details. When using navigation arrows, it doesn't load the preview of the task.

0x4007:
> It seems that external organization task items are broken. When clicking on the item, it redirects to the task instead of opening the popup with the details. When using navigation arrows, it doesn't load the preview of the task.

It's because there is no issue body. That's by design.

scripts/dev.ts
0x4007: Because LLMs suck at dealing with servers.

0x4007:
> Could you explain the reason to still use `esbuild` here?

In this repo, esbuild is used via Deno for bundling assets, CSS processing, PWA manifest generation, and injecting env defines (see build/esbuild-build.ts:18-43), which Deno’s built-in bundling doesn’t fully replace

gentlementlegen:
If we move to Deno, the environment should be handled with Deno.env, and Deno has `bundle` or build already implemented so I believe it should be moved there instead

gentlementlegen:
> > It seems that external organization task items are broken. When clicking on the item, it redirects to the task instead of opening the popup with the details. When using navigation arrows, it doesn't load the preview of the task.
> 
> It's because there is no issue body. That's by design.

I am pretty sure it used to display the external organizations tasks but okay I truest you. However, when using keyboard navigation, the title changes but the body seems to remain the same.

0x4007:
> > > It seems that external organization task items are broken. When clicking on the item, it redirects to the task instead of opening the popup with the details. When using navigation arrows, it doesn't load the preview of the task.
> > 
> > 
> > It's because there is no issue body. That's by design.
> 
> I am pretty sure it used to display the external organizations tasks but okay I truest you. However, when using keyboard navigation, the title changes but the body seems to remain the same.

Then this is out of scope I haven't noticed it before, but its also extremely rare that we create tasks with no spec.

0x4007:
> If we move to Deno, the environment should be handled with Deno.env, and Deno has `bundle` or build already implemented so I believe it should be moved there instead

> • - Thanks — agreed on both points. We’ve moved config to Deno.env and load .env in dev via std/dotenv. See server.ts:2,
>     server.ts:5, functions/issue-scraper.ts:131, and the build script using Deno.env.get for define injection build/esbuild-
>     build.ts:1, build/esbuild-build.ts:16, build/esbuild-build.ts:35. Docs reflect this: README.md:23.
>   - For bundling: we’re deliberately using esbuild through Deno (deno run -A npm:esbuild) because we need CSS/asset handling,
>     loaders (fonts/images as data URLs), a couple of small plugins (PWA manifest rewrite and grayscale inversion), and
>     build‑time constant injection for the client. deno bundle focuses on TS/JS and doesn’t provide a plugin system or CSS/
>     asset pipeline, so we’d end up needing a second tool anyway. This keeps us Deno‑only while meeting the current requirements
>     (policy noted in AGENTS.md:29).
>   - Happy to revisit deno bundle once first‑party tooling covers CSS/assets and define injection; until then I’d keep
>     esbuild‑in‑Deno for a single‑pass, predictable build.

deno.json
0x4007: Addressed: removed redundant --unstable-kv flags. deno.json tasks now rely on "unstable": ["kv"]. Updated: deno.json:3, deno.json:13.

scripts/dev.ts
0x4007: Kept the dev runner. It starts server+esbuild concurrently, auto-picks a free port when PORT is unset, and cleans up both children on SIGINT/SIGTERM (useful in CI/agent runs). Expanded header comments in scripts/dev.ts.

0x4007: Added a concise rationale above; leaving the helper in for reliable local + CI dev flow.

server.ts
0x4007: Refactored duplicate onListen blocks into a single helper (onListenCommon) to write port/pid files and log the URL. See server.ts.

src/home/fetch-github/fetch-and-display-previews.ts
0x4007: Replaced window usage with globalThis (null-safe) in fetch-and-display-previews.ts: location/history/pathname/redirect. This makes the module safe if ever imported in Deno contexts.

src/home/getters/get-github-access-token.ts
0x4007: It is used: src/home/authentication.ts:15, src/home/task-manager.ts:43, src/home/rendering/display-github-user-information.ts:36. Keeping as-is.

.github/workflows/build.yml
```diff
@@ -18,25 +18,27 @@ jobs:
...
 
       - name: Build
         run: |
-          yarn
-          yarn build
-        env: # Set environment variables for the build
-          SUPABASE_URL: "https://wfzpewmlyiozupulbuur.supabase.co"
-          SUPABASE_ANON_KEY: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6IndmenBld21seWlvenVwdWxidXVyIiwicm9sZSI6ImFub24iLCJpYXQiOjE2OTU2NzQzMzksImV4cCI6MjAxMTI1MDMzOX0.SKIL3Q0NOBaMehH0ekFspwgcu3afp3Dl9EDzPqs1nKs"
+          export NODE_ENV="${NODE_ENV:-production}"
+          export SUPABASE_URL="${SUPABASE_URL:-https://placeholder.supabase.co}"
```
gentlementlegen: Shouldn't the build fail instead of having defaults here?

deno.json
```diff
@@ -0,0 +1,42 @@
...
+    "serve:15s": "deno run --allow-run --allow-read --allow-write --allow-env --allow-net scripts/serve_15s.ts",
+    "serve:bg": "deno run --allow-run --allow-read --allow-write --allow-env --allow-net scripts/serve_bg.ts",
+    "serve:stop": "deno run --allow-read --allow-write scripts/serve_stop.ts",
+    "serve:status": "deno run --allow-net --allow-read --allow-env scripts/serve_status.ts",
+    "serve:wait": "deno run --allow-run --allow-net --allow-read --allow-write --allow-env scripts/serve_wait.ts",
+    "build": "deno run -A build/esbuild-build.ts",
+    "watch": "deno run -A build/esbuild-watch.ts",
+    "dev:bg": "deno run --allow-run --allow-read --allow-write --allow-env --allow-net scripts/dev_bg.ts",
+    "dev:stop": "deno run --allow-read --allow-write scripts/dev_stop.ts",
+    "dev": "deno run --unstable-kv -A scripts/dev.ts",
```
gentlementlegen:
```suggestion
    "dev": "deno run -A scripts/dev.ts",
```

scripts/dev.ts
```diff
@@ -0,0 +1,54 @@
...
+  } catch (_) {}
+  try {
+    watchProc.kill("SIGTERM");
+  } catch (_) {}
+  // Give children a brief moment to exit
+  setTimeout(() => Deno.exit(0), 50);
+}
+
+Deno.addSignalListener("SIGINT", () => gracefulShutdown("SIGINT"));
+Deno.addSignalListener("SIGTERM", () => gracefulShutdown("SIGTERM"));
```
gentlementlegen: `SIGTERM` is not cross-platform, you only have only ctrl-c (SIGINT), ctrl-break (SIGBREAK), and ctrl-close (SIGUP), but no SIGTERM

gentlementlegen:
```suggestion
```

ubiquity-ubiquibot (approved):
I don't like having so much technical debt because LLMs are stupid, but I guess we need the FE to work again

gentlementlegen (approved):
I don't like having so much technical debt because LLMs are stupid, but I guess we need the FE to work again

scripts/dev.ts
0x4007: This works for everywhere. We clearly added the event listener to both.

deno.json
0x4007: No we are using an unstable KV API.

